PR #1, opened 2023-04-18
Treat RHEL major version comparison specially

comps:
Compare majors between themselves, ignoring RPM NVR comparison.

```
rhel > 7    # True on rhel-8, False on rhel-7.9
rhel > 7.0  # True on rhel-8, True on rhel-7.9
```


mildas:
LGTM

PR #2, opened 2023-04-19
Implement an auto-waiving logic for expected failures

comps:
Adding a configurable waiving logic for test results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,14 @@ tmt \
             report -h html
 ```
 
+## Waiving failed results
+
+In this context, "to waive" means to label a failing result as known-bad,
+something we have seen before and expect to fail.
+
+Read [WAIVES.md](WAIVES.md) to see where/how you can set up rules to
+automatically waive failures.
+
 ## Workarounds
 
 (TODO: Find a better place for this?)
--- a/README.md
+++ b/README.md
@@ -31,10 +31,6 @@ on Red Hat Enterprise Linux.
 
 (TODO: Probably document this on a better place.)
 
-- `CONTEST_SILENT`
-  - set to `1` to verbosely report only genuine oscap rule failures
-  - other results are unaffected (for now)
-
 - `CONTEST_DATASTREAMS`
   - alternate location for `/usr/share/xml/scap/ssg/content`
 - `CONTEST_PLAYBOOKS`
--- a/WAIVES.md
+++ b/WAIVES.md
@@ -1,0 +1,133 @@
+# Waiving known failures
+
+## Custom file format
+
+The [waives file](conf/waives) uses a custom file format to specify when
+a failure is expected.  
+This format consists of so-called "sections", optionally separated by any
+number of empty lines. Any lines beginning with `#` are skipped.
+
+```
+<section>
+
+# some comment
+<section>
+
+...
+```
+
+Each section consists of two parts:
+
+- a list of result names, as regular expressions
+- a whitespace-indented block with a python code expression
+
+```python
+/some/result/name
+/some/other/.*/result
+    True if some_condition else \
+        True if some_other_condition else False
+
+/some/more/test/results
+    rhel < 9 and ssg < '0.1.66'
+```
+
+When a new failing result is to be reported, the sections are evaluated
+sequentially, in order, **from the top**.
+
+Within each section:  
+If at least one of the consecutive regular expressions matches the result name,
+the python expression is evaluated.  
+If that expression also returns `True`, the result is considered waived and
+the traversal stops for that one result (no further sections are tried).
+
+If either the regular expressions don't match, or the python expression returns
+False, the section is skipped and a next one is considered.
+
+If none of the sections match, no waiving happens and the result remains
+unchanged.
+
+## Regular expressions
+
+The regexps use a pythonic syntax, and are matched using `re.fullmatch()`,
+meaning they always match the full result name, not a substring.
+
+This is to easily and naturally support specifying result names as-is,
+without any regular expression characters.
+
+Ie. this behaves as expected, always matching only on the full name:
+
+```
+/some/result/name
+```
+
+If you need to match only a substring (prefix), emulating `re.match()`,
+use wildcards:
+
+```
+/some/result/name.*
+```
+
+Or `re.search()` behavior using:
+
+```
+.*/some/result/name.*
+```
+
+## Not just a `fail`
+
+The waiving logic is actually evaluated for all of `pass`, `fail` and `error`.
+The `pass` is there to catch any waive matches that suddenly started `pass`-ing,
+and the `error` allows us to waive some infrastructure `error`s as well.
+
+The waiving file format above therefore serves as a generic "matching logic",
+and it's up to the library code to interpret what a match means.
+
+- no match = no change
+- match + pass = error (unexpected pass)
+- match + fail = warn (waived fail)
+- match + error = warn (waived error)
+
+## Matching and globals
+
+The python code block must always be one expression, not a freeform python
+module code.
+
+The expression has these globals available:
+
+- `status` - the original result status (`pass`, `fail`, `error`, etc.)
+- `name` - the result (test) name, empty string if unspecified
+- `note` - an optional note associated with the result, or empty string
+- `rhel` - an object capable of RHEL version comparison, see
+  [versions.rhel](lib/versions.py)
+- `oscap` - an object capable of `openscap-scanner` RPM version comparisons,
+  see [versions.oscap](lib/versions.py)
+- `ssg` - same, but for the `scap-security-guide` RPM
+- `env` - environment variable retrieval function, same as `os.environ.get()`
+- `Match` - a class for complex waive results (see below)
+
+The version comparison objects also support a boolean evaluation, with
+`bool(rhel)` returning `False` if not running on RHEL, and ie. `bool(oscap)`
+returning `False` if the RPM is not installed.
+
+```python
+/some/result/name
+    (rhel < 7.9 and 'some thing' in note) or \
+    (rhel < 8.4 and oscap < '0.1.66')
+
+/some/other/result
+    rhel == 8 and 'some thing' in note and env('INFRA') == 'jenkins'
+```
+
+### Expecting both `pass` and `fail`/`error`
+
+Sometimes, failures or errors are not reliable and happen only occassionally.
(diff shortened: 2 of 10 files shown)


comps:
I believe this is ready for review now.

Some notes:

The `CONTEST_SILENT` removal is there because I found it much easier to have all the results available to me when creating a waive rule, so I wouldn't have to rerun all tests multiple times to verify it really catches everything and nothing more. Having access to `pass` results helped with that. With https://github.com/teemtee/tmt/issues/2009, having thousands of results shouldn't be a big problem. For Beaker, there's a compromise of reporting only failures (making Beaker results less useful for waive debugging), see the commit message.

Then there's failure handling - `CONTEST_SILENT` was problematic even from a code POV as it decided what should/shouldn't be logged based on the original status (before waiving, in `oscap.report_from_verbose()`) as waiving is done in the result reporting function itself, later. I didn't want to move the silencing to apply to **all** results because ie. we have to report an overall `pass` for the test itself. Getting rid of `CONTEST_SILENT` spared me some ugly code (duplicating waiving calls).

That also meant I had to move failure counting (whether we should exit with 0 or non-0, mainly for Beaker) into the result reporting module (getting rid of the `failed` return from `oscap.report_from_verbose()`) - I'm not super happy with that as it breaks one of the style rules by keeping state (`failed_count` / `errored_count`) in the non-test module, but there was no nicer way of doing this "singleton-style" logic. Having a proper class with a single instance would just add one layer of needless abstraction.  
On the flipside, it simplified the test a bit with `report_and_exit()`.

comps:
It would be great if you could also help clarify some of the TODOs in `conf/waives` as those were seemingly untracked by https://src.fedoraproject.org/tests/scap-security-guide/blob/main/f/Library/scap-results/rule_waivers

README.md
```diff
@@ -76,6 +72,14 @@ tmt \
             report -h html
 ```
 
+## Waiving failed results
+
+In this context, "to waive" means to discard a failing result as known-bad,
```
mildas, line 77: `to discard` is not correct wording. Results won't be discarded, they are not going to cause overall fail, but still will be shown as WARN.

WAIVES.md
```diff
@@ -0,0 +1,136 @@
...
+# some comment
+<section>
+
+...
+```
+
+Each section consists of two parts:
+
+- a list of result names, as regular expressions
+- a whitespace-indented block with a python code expression
```
mildas, line 22: Are we talking about arbitrary indented block, or it must be 4spaces per level?

conf/waives
```diff
@@ -0,0 +1,53 @@
+# TODO: unknown, but maybe expected as the rule description says that
+#   "The remediation for this rule uses firewall-cmd and nmcli tools.
+#    Therefore, it will only be executed if firewalld and NetworkManager
+#    services are running."
+# and since these are not running in an Anaconda %post environment,
+# maybe the remediation failing is expected .. ?
+# - see also https://github.com/ComplianceAsCode/content/issues/1639
```
mildas, line 7: Very old issue, the rule has been reworked.
Use https://github.com/ComplianceAsCode/content/issues/9746

conf/waives
```diff
@@ -0,0 +1,53 @@
...
+/hardening/oscap/[^/]+/no_tmux_in_shells
+/hardening/oscap/[^/]+/configure_usbguard_auditbackend
+    rhel == 9 or rhel == 8
+/hardening/oscap/[^/]+/configure_bashrc_tmux
+/hardening/oscap/stig(_gui)?/postfix_prevent_unrestricted_relay
+    rhel == 8
+
+# TODO: old test waives say this is also
+# https://github.com/OpenSCAP/openscap/issues/1880
+# but that seems to be incorrect as (unlike the others),
```
mildas, line 23: The problem here is combination with `enable_authselect` rule. `accounts_password_pam_retry` passes during initial scan, `enable_authselect` forcibly sets authselect profile and that changes `accounts_password_pam_retry` result.

README.md
comps: This is a high-level description from an overall process workflow PoV. "To discard" as in "to render uninteresting". Maybe I can try rephrasing it differently.

WAIVES.md
comps: Any arbitrary indented code block, as passed through `textwrap.dedent()`, eg. the largest whitespace indent shared by all lines in that block will be removed.
```
      a b
        c d
```
to
```
a b
  c d
```

WAIVES.md
```diff
@@ -0,0 +1,136 @@
...
+- `env` - environment variable retrieval function, same as `os.environ.get()`
+- `Match` - a class for complex waive results (see below)
+
+The version comparison objects also support a boolean evaluation, with
+`bool(rhel)` returning `False` if not running on RHEL, and ie. `bool(oscap)`
+returning `False` if the RPM is not installed.
+
+```python
+/some/result/name
+    (
```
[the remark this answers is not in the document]
comps: I was just trying to show off different ways this could be done, to make the reader understand that `\` is not required for multiline code.

But maybe we shouldn't even do that and just always use the same syntax?

conf/waives
comps: Thanks, will change the comment. The curious part is that this still seems to be failing even with current upstream content from git.

conf/waives
comps: Ok, I'll mention this in the comment. Thanks.

WAIVES.md
mildas: No need to clarify it in doc.

WAIVES.md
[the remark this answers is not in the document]
mildas: > But maybe we shouldn't even do that and just always use the same syntax?

Yes, let's always use the same syntax.

conf/waives
mildas: Well, the issue is closed because it is not a problem in content. The problem is in Anaconda environment and we can't affect that, so it will fail there most likely forever.

lib/waive.py
```diff
@@ -0,0 +1,215 @@
...
+            if line:
+                state = 'reading_regex'
+                lines.pushback()
+                continue
+
+        elif state == 'reading_regex':
+            if not line:
+                raise WaiveParseError(lines, "unexpected empty line between regexes")
+
+            if line.startswith((' ', '\t')):
```
comps, line 87: This probably should be a list, to be consistent with other similar enumerations.

lib/waive.py
```diff
@@ -0,0 +1,215 @@
...
+                lines.pushback()
+                continue
+
+            try:
+                regexes.add(re.compile(line))
+            except re.error as e:
+                raise WaiveParseError(lines, f"regex failed: {e}")
+
+        elif state == 'reading_python':
+            if not line.startswith((' ', '\t')):
```
comps, line 100: List here as well.

lib/waive.py
```diff
@@ -0,0 +1,215 @@
...
+    for line in lines:
+        if line.startswith('#'):
+            continue
+        line = line.rstrip('\n')
+
+        if state == 'skipping_empty_lines':
+            if line:
+                state = 'reading_regex'
+                lines.pushback()
+                continue
```
mildas, line 81: I'd prefer making the "empty skipping" more obvious there:
```suggestion
            if not line:
                continue
            else:
                state = 'reading_regex'
                lines.pushback()
```

lib/waive.py
```diff
@@ -0,0 +1,215 @@
...
+        elif state == 'reading_python':
+            if not line.startswith((' ', '\t')):
+                sections.append((regexes, _compile_eval(lines, python_code)))
+                regexes = set()
+                python_code = ''
+                state = 'skipping_empty_lines'
+                lines.pushback()
+                continue
+
+            python_code += f'{line}\n'
```
mildas, line 108: Similar suggestion as above, basically trying to match expected input first
```suggestion
            if line.startswith((' ', '\t')):
                python_code += f'{line}\n'
                continue
            else:
                sections.append((regexes, _compile_eval(lines, python_code)))
                regexes = set()
                python_code = ''
                state = 'skipping_empty_lines'
                lines.pushback()
```

lib/waive.py
```diff
@@ -0,0 +1,215 @@
...
+                    raise WaiveParseError(lines, "python block without a preceding regexp")
+                state = 'reading_python'
+                lines.pushback()
+                continue
+
+            try:
+                regexes.add(re.compile(line))
+            except re.error as e:
+                raise WaiveParseError(lines, f"regex failed: {e}")
+
```
mildas, line 98: Just a suggestion to make the logic there more obvious. It took me some time to get the point in the original code so I tried to come up with something simpler, but it still doesn't feel right :/
```suggestion
            if not line.startswith((' ', '\t')):
                try:
                    regexes.add(re.compile(line))
                except re.error as e:
                    raise WaiveParseError(lines, f"regex failed: {e}")
                continue
            else:
                if regexes:
                    state = 'reading_python'
                    lines.pushback()
                else:
                    raise WaiveParseError(lines, "python block without a preceding regexp")
```

mildas (commented):
Most of the code is well readable, and the flow is pretty clear. But I had a problem with parsing waiving file. Not that I didn't get the logic, but it took me significant amount of time to fully understand it.
I've proposed some suggestions. They aren't ideal, just a different point of view for a discussion.

conf/waives
```diff
@@ -0,0 +1,53 @@
...
+/hardening/anaconda/e8/rpm_verify_(ownership|permissions)
+    rhel == 7
+
+# bz1825810 or maybe bz1929805
+# can be reproduced reliably both via anaconda and oscap CLI,
+# but apparently we don't really care about old releases
+/hardening/[^/]+/stig_gui/sysctl_net_ipv4_ip_forward
+    rhel == 7
+
+# TODO: something new? .. RHEL-8 on e8 and ism_o
```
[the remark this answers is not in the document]
comps: I'll leave it as TODO for now, investigation can probably be done elsewhere. It fails consistently, otherwise I'd have gotten errors (pass, but matched).

mildas (approved):

comps:
Just fixing an extra whitespace discovered by `flake8`. We should really implement https://github.com/RHSecurityCompliance/contest/issues/5

comps:
This PR passed a full run of all current tests on RHEL 7/8/9, so given the approval, I'm merging this. Thanks for the review!

PR #3, opened 2023-04-20
allow more special characters in make_printable

comps:
None of these should cause issues as Exception strings, or YAML strings.


mildas:
LGTM

PR #8, opened 2023-04-26
Use shipped/upstream content (xml, kickstarts) directly

comps:
(Instead of using whatever is in the compose.)

This makes
* `oscap` use a copied datastream XML directly
* `anaconda` to use a HTTP-server-supplied datastream XML
* (future) ansible test to use a playbook directly

Essentially this enables us to source all the content pieces from wherever we want, rather than relying on them being available in the compose.

It also makes the `/hardening/anaconda` test already use content-provided kickstarts, rather than our own ones.

I'm not super proud of the `firewalld` workaround logic, but I think it's GEFN until we figure out all the environments we want to run in. Minimal install (ie. `easy-virt-install` kickstart) uses firewalld enabled by default, Beaker disables it for its installs, the zone used by libvirt differs between releases, etc.

Probably go through this commit-by-commit.

There's a few TODOs left in the code comments that I plan to resolve soon (1-2 days) after I get a reply from some package groups (compositions) experts and submit a CaC/content PR for potentially changing `%packages` sections of kickstarts, after which we may not need to clear them out.

Additional changes (testing all profiles via `Server with GUI`) will follow later, and are currently out of scope for this PR.

### DIFF
--- a/hardening/anaconda/test.py
+++ b/hardening/anaconda/test.py
@@ -12,38 +12,42 @@ import versions
 
 virt.setup_host()
 
-profile = os.environ['PROFILE']
-profile = f'xccdf_org.ssgproject.content_profile_{profile}'
-
 g = virt.Guest()
 
-ks = virt.Kickstart()
+profile = os.environ['PROFILE']
 
-# remediate using Anaconda's oscap addon
-oscap_conf = {
-    'content-type': 'scap-security-guide',
-    'profile': profile,
-}
-ks.add_oscap(oscap_conf)
+# use kickstart from content, not ours
+ks = virt.translate_ssg_kickstart(profile)
+
+profile = f'xccdf_org.ssgproject.content_profile_{profile}'
 
 ks.add_post('chage -d 99999 root')
 
 if profile.endswith('_gui'):
     ks.add_package_group('Server with GUI')
 
-g.install(kickstart=ks)
+oscap_conf = {
+    'content-type': 'datastream',
+    'content-url': f'http://{virt.NETWORK_HOST}:8088/contest-ds.xml',
+    'profile': profile,
+}
+ks.add_oscap(oscap_conf)
+
+# host a HTTP server with a datastream and let the guest download it
+virt.firewalld_allow_port(8088)
+srv = util.BackgroundHTTPServer(virt.NETWORK_HOST, 8088)
+srv.add_file(util.get_datastream(), 'contest-ds.xml')
+with srv:
+    g.install(kickstart=ks)
 
 with g.booted():
-    # copy our datastream to the guest
-    g.copy_to(util.get_datastream(), 'contest-ds.xml')
-
     # old RHEL-7 oscap mixes errors into --progress rule names without a newline
     verbose = '--verbose INFO' if versions.oscap >= 1.3 else ''
     redir = '2>&1' if versions.oscap >= 1.3 else ''
 
     # scan the remediated system
     proc, lines = g.ssh_stream(f'oscap xccdf eval {verbose} --profile {profile} --progress '
-                               f'--report report.html contest-ds.xml {redir}')
+                               f'--report report.html /root/openscap_data/contest-ds.xml {redir}')
     failed = oscap.report_from_verbose(lines)
     if proc.returncode not in [0,2]:
         raise RuntimeError("post-reboot oscap failed unexpectedly")
--- a/hardening/anaconda/test.py
+++ b/hardening/anaconda/test.py
@@ -3,6 +3,7 @@
 import os
 import sys
 
+import util
 import results
 import virt
 import oscap
@@ -33,13 +34,16 @@ if profile.endswith('_gui'):
 g.install(kickstart=ks)
 
 with g.booted():
+    # copy our datastream to the guest
+    g.copy_to(util.get_datastream(), 'contest-ds.xml')
+
     # old RHEL-7 oscap mixes errors into --progress rule names without a newline
     verbose = '--verbose INFO' if versions.oscap >= 1.3 else ''
     redir = '2>&1' if versions.oscap >= 1.3 else ''
 
     # scan the remediated system
     proc, lines = g.ssh_stream(f'oscap xccdf eval {verbose} --profile {profile} --progress '
-                               f'--report report.html {oscap.DATASTREAM} {redir}')
+                               f'--report report.html contest-ds.xml {redir}')
     failed = oscap.report_from_verbose(lines)
     if proc.returncode not in [0,2]:
         raise RuntimeError("post-reboot oscap failed unexpectedly")
--- a/hardening/anaconda/test.py
+++ b/hardening/anaconda/test.py
@@ -34,7 +34,6 @@ oscap_conf = {
 ks.add_oscap(oscap_conf)
 
 # host a HTTP server with a datastream and let the guest download it
-virt.firewalld_allow_port(8088)
 srv = util.BackgroundHTTPServer(virt.NETWORK_HOST, 8088)
 srv.add_file(util.get_datastream(), 'contest-ds.xml')
 with srv:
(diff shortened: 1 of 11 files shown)


lib/virt.py
```diff
@@ -819,6 +827,51 @@ def ssh_keygen(path):
...
+            if re.match('^rootpw ', line):
+                line = f'rootpw {GUEST_LOGIN_PASS}'
+
+            # don't hardcode interface name
+            if re.match('^network ', line):
+                line = re.sub(' --device[^ ]+', '', line)
+
+            # shrink some unnecessary partitions
+            if re.match('^logvol / ', line):
+                line = re.sub('--size=[^ ]+', '--size=4000', line)
```
comps, line 850: Maybe I should check whether the original isn't `>4000` to make sure we're not enlarging it and going over the limit.

hardening/anaconda/test.py
```diff
@@ -33,13 +34,16 @@
 g.install(kickstart=ks)
 
 with g.booted():
+    # copy our datastream to the guest
+    g.copy_to(util.get_datastream(), 'contest-ds.xml')
```
mildas: I'd prefer keeping filename the same, `ssg-rhel{versions.rhel.major}-ds.xml`. Because I'm used to that. But it's not blocker, just my 2cents.

mildas: Applies also to `oscap` test.

lib/virt.py
mildas: Dunno, if there is single value we could use for all profiles, as some have `/` <4000 - ANSSI https://github.com/ComplianceAsCode/content/blob/master/products/rhel8/kickstart/ssg-rhel8-anssi_bp28_high-ks.cfg#L99
and some have it `>>4000` as OSPP https://github.com/ComplianceAsCode/content/blob/master/products/rhel8/kickstart/ssg-rhel8-ospp-ks.cfg#L96

lib/virt.py
```diff
@@ -819,6 +827,51 @@ def ssh_keygen(path):
...
+    with open(util.get_kickstart(profile)) as f:
+        for line in f:
+            line = line.rstrip('\n')
+
+            # use our own password
+            if re.match('^rootpw ', line):
+                line = f'rootpw {GUEST_LOGIN_PASS}'
+
+            # don't hardcode interface name
+            if re.match('^network ', line):
```
mildas, line 845: As you are cycling over lines, there will be single match every time. Thus, to make it obvious, use `elif` statement here and further.

comps: I guess I saw the matches as more independent (easier to add/remove when there's no `elif`), but I can make that optimization, sure.

lib/virt.py
```diff
@@ -820,6 +827,51 @@ def ssh_keygen(path):
...
+    ks_text = re.sub(r'\n%addon .+?_oscap\n.+?\n%end[^\n]*', '', ks_text, flags=re.DOTALL)
+
+    # remove %packages, until it becomes consistent everywhere
+    # TODO: upstream content PR + add it to README
+    ks_text = re.sub(r'\n%packages.*?\n%end[^\n]*', '', ks_text, flags=re.DOTALL)
+
+    return Kickstart(template=ks_text)
+
+
+def firewalld_allow_port(port):
```
mildas, line 866: Wouldn't `utils.py` be a better fit? To move it closer to HTTP Server as those two functions are related (and most likely will be the only ones).
Maybe even use within HTTP Server? Just food for thought, opening/closing the ports with HTTP server.

comps: I did consider that, but because it explicitly involved the `libvirt` zone, I put it in `virt.py`. I also considered putting it into the HTTP server class, but once again decided against it because of the `libvirt` zone.

Another thought I had that we could do is to generalize it as "add the port to all firewalld zones" or even "check for iptables/nftables and add a transient rule manually", ignoring firewalld and its zones, making it generic (non-libvirt), allowing it to be placed in the util/HTTP code.

Since you pointed this out, I think I'll do it. I also don't want any more code in `virt.py` than necessary.

hardening/anaconda/test.py
mildas: Okay, make sense.

lib/virt.py
comps: Hmm, perhaps that wouldn't do it.

`iptables` could simply insert an `ACCEPT` rule into `INPUT`, but `nftables` is much more complex due to its nature of multiple tables, each having multiple chains, and `accept`-ing in a higher priority one doesn't fully accept the packet, it just makes it go onto other filters.  
So we would need to insert it into every chain of every table.

At that point, we might as well just disable the firewall and restore it after the HTTP server shuts down. But that would mean using blocks of empty code for `iptables-restore` and storing the original `nft` ruleset, maybe breaking `libvirt` itself.

... So I may go the simpler way and just use libvirt's `nwfilter`, which is basically a XML language for writing firewall rules for virtual machines, and make all this libvirt-specific anyway. At least there won't be arbitrary `firewalld` stuff in `virt.py`.

comps: Well, that way also isn't too great, as nwfilter is a whole new socket/service that's disabled by default, and the language would need another XML block in `virt.py`.

I think I'll just do it on the firewalld level, "list all zones, parse it, add rule to every zone, prepare cleanup removing them".

comps:
I have moved the firewall code to `BackgroundHTTPServer` inside `util.py`, changed multiple `if` into `elif` when modifying upstream content kickstarts, and added some README documentation (will fill CaC/content PR tomorrow).

I also kept the sanitized `content-ds.xml` name, but at least unified it across the `anaconda` and `oscap` tests.

comps:
So one more batch of changes and a switch out of Draft.

On the content side, I created https://github.com/ComplianceAsCode/content/pull/10499 and while working on it, I found out that many of the kickstart hacks were not actually needed, so I cleaned up and commented `translate_ssg_kickstart()`.  
The 4000 MB thing for `/` is gone.

I also discovered that `cui` and `pci-dss` don't have RHEL-7 kickstarts, so I disabled them on RHEL-7.

Finally, sorry for increasing the scope of this by adding the `/plan/upstream` stuff, but I needed some way to actually test all this code on the kickstarts I fixed in https://github.com/ComplianceAsCode/content/pull/10499.  
The `README` stuff is there just as a sidenote (for now) and its refinement is probably out of scope here.

(Don't ask me about the `prepare+` TMT magic, I don't know. It's something I cobbled up together from 2 hours of experiments, and it seems to be working thanks the the usage of YAML lists.)

Let me know if you have any other comments/concerns.

mildas:
> I also discovered that cui and pci-dss don't have RHEL-7 kickstarts, so I disabled them on RHEL-7.

Actually, PCI DSS kickstart is there, it just doesn't have unified format there and moreover it with GUI packages - https://github.com/ComplianceAsCode/content/blob/master/products/rhel7/kickstart/ssg-rhel7-pci-dss-server-with-gui-oaa-ks.cfg so we can ignore Minimal variant there and test only this one.

Other than that, I don't have any other objections. I will try it with brew build and then it's ready for merge.

comps:
Fixed Beaker's brew build task breaking our code - `/distribution/install/brew-build` actually sets up a local `file://`-based YUM repo, which got copied to the guest and when Anaconda didn't find the filepath in its installation environment, it broke down.  
So the easy fix is to ignore any `file://` repositories on the host.

> Actually, PCI DSS kickstart is there, it just doesn't have unified format there and moreover it with GUI packages - https://github.com/ComplianceAsCode/content/blob/master/products/rhel7/kickstart/ssg-rhel7-pci-dss-server-with-gui-oaa-ks.cfg so we can ignore Minimal variant there and test only this one.

It will be an interesting discussion in another PR (one which is going to add `Server with GUI` variants for all profiles), as there are a few options to consider, ie.

1. just skipping special cases like these on RHEL-7, as a "not worth the hacks it would introduce" kind of reasoning ... (not because it's GUI only, but because it has a non-standard name)
1. adding some hack to `get_kickstart()` that would check for `rhel == 7 and profile == 'pci-dss'` and return a special hardcoded filename
1. adding some environment variable that would do the same (override kickstart file name, not just directory like `CONTEST_KICKSTARTS` does)
1. fixing it in content upstream right now, and skipping it (like first point) until we release new content

PR #11, opened 2023-05-05
run all oscap/anaconda tests on "Server with GUI" too

comps:
The `2h` duration is there because the extra packages do add +20-30min of time, sometimes more.

### DIFF
--- a/hardening/anaconda/with-gui.fmf
+++ b/hardening/anaconda/with-gui.fmf
@@ -1,0 +1,83 @@
+environment+:
+    USE_SERVER_WITH_GUI: 1
+duration: 2h
+
+/anssi_bp28_high:
+    environment+:
+        PROFILE: anssi_bp28_high
+    adjust:
+        when: distro < rhel-8
+        enabled: false
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/anssi_bp28_high
+
+/anssi_nt28_high:
+    environment+:
+        PROFILE: anssi_nt28_high
+    adjust:
+        when: distro > rhel-7
+        enabled: false
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/anssi_nt28_high
+
+/cis:
+    environment+:
+        PROFILE: cis
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/cis
+
+/cis_server_l1:
+    environment+:
+        PROFILE: cis_server_l1
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/cis_server_l1
+
+/cis_workstation_l1:
+    environment+:
+        PROFILE: cis_workstation_l1
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/cis_workstation_l1
+
+/cis_workstation_l2:
+    environment+:
+        PROFILE: cis_workstation_l2
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/cis_workstation_l2
+
+/cui:
+    environment+:
+        PROFILE: cui
+    adjust:
+        when: distro < rhel-8
+        enabled: false
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/cui
+
+/e8:
+    environment+:
+        PROFILE: e8
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/e8
+
+/hipaa:
+    environment+:
+        PROFILE: hipaa
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/hipaa
+
+/ism_o:
+    environment+:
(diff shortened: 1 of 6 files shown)


mildas:
> To me, the with-gui case is a special case, and we could potentially have more like that, but the minimal installation is the baseline standard case

That's an interesting point. I always perceived them on same importance level. But Minimal as baseline and GUI as something extra is valid.

If we could come with ideas for more special cases as GUI (e.g. common machine configuration), then leaving code as it is now is fine to me.
Basically, I'd like to know that GUI won't be the only one with the extra hierarchy level. If it would be the only one, then we should consider different approaches, e.g. everything in `main.fmf`:
```
/hardening/oscap/ospp
/hardening/oscap/ospp (gui)
```

comps:
> If we could come with ideas for more special cases as GUI (e.g. common machine configuration)

Yeah, I can't think of anything right now. Multiple architectures maybe, but that would be done on a higher level.

> If it would be the only one, then we should consider different approaches, e.g. everything in `main.fmf`:

Okay, that should be possible.

It would look something like

```
...
environment+:
    PYTHONPATH: ../../lib
duration: 1h
extra-hardware: |
    keyvalue = HVM=1
    hostrequire = memory>=4800
...
/cis:
    environment+:
        PROFILE: cis
...
/with-gui:
    environment+:
        USE_SERVER_WITH_GUI: 1
    duration: 2h
    ...
    /cis:
        environment+:
            PROFILE: cis
    ...
```

cortesana:
My preference on this is to spread horizontally, i.e.:
```
/annocheck
/aslr-bits
/kernel/cmdline
...
```
rather than vertically, i.e.:
```
other/annocheck
other/aslr-bits
/kernel/cmdline
...
```
If further down the road there is an increase in the number of special cases on the same hierarchy and exploring the tests becomes a problem, this could easily be overcome by further classification of the cases into meaningful groups.

> Basically, I'd like to know that GUI won't be the only one with the extra hierarchy level. If it would be the only one, then we should consider different approaches, e.g. everything in `main.fmf`

I prefer the extra hierarchy level over everything being in `main.fmf`. @mildas What is the consideration behind not wanting GUI to be the only one with the extra hierarchy level?

mildas:
> What is the consideration behind not wanting GUI to be the only one with the extra hierarchy level?

Then we would have 2 very similar tests on 2 different hierarchy levels. Basically, an additional hierarchy level only because of extra packages that come with GUI. Roots of it probably come from my previous understanding of Minimal and GUI installations - considering them still as equivalent.

And at least knowing, there will be more test cases in the extra hierarchy level, would convince me it makes sense to have it. It would be more understandable also for others to see the difference.

cortesana:
Thanks, @mildas!

mildas (changes requested):
CIS Server Lever 1 (`cis_server_l1`) and Level 2 (`cis`) shouldn't be enabled on GUI. OSPP and CUI can be tested with GUI only on RHEL9=<

See docs
- RHEL8 https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/security_hardening/index#ref_profiles-not-compatible-with-server-with-gui_deploying-systems-that-are-compliant-with-a-security-profile-immediately-after-an-installation
- RHEL9 https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html-single/security_hardening/index#ref_profiles-not-compatible-with-server-with-gui_deploying-systems-that-are-compliant-with-a-security-profile-immediately-after-an-installation

hardening/oscap/main.fmf
```diff
@@ -83,6 +85,7 @@ extra-hardware: |
     adjust:
         when: distro < rhel-8
         enabled: false
+        because: exists only on RHEL-7
```
mildas, line 88: `doesn't exist on RHEL-7`

hardening/oscap/with-gui.fmf
```diff
@@ -0,0 +1,113 @@
...
+        when: distro == rhel-7
+        enabled: false
+        because: doesn't exist on RHEL-7
+    extra-summary: /CoreOS/scap-security-guide/hardening/oscap/with-gui/ism_o
+
+/ospp:
+    environment+:
+        PROFILE: ospp
+    adjust:
+        when: distro <= rhel-8
```
mildas, line 87: `distro == rhel-8`

hardening/anaconda/with-gui.fmf
```diff
@@ -0,0 +1,123 @@
...
+        when: distro == rhel-7
+        enabled: false
+        because: doesn't exist on RHEL-7
+    extra-summary: /CoreOS/scap-security-guide/hardening/anaconda/with-gui/ism_o
+
+/ospp:
+    environment+:
+        PROFILE: ospp
+    adjust:
+        when: distro <= rhel-8
```
mildas, line 90: `distro == rhel-8`

mildas (changes requested):

mildas (approved):
We have decided not to test OSPP with GUI on RHEL7.
Everything else was already LGTM.

PR #18, opened 2023-05-24
Add a `/hardening/ansible` test, with gui too

comps:
Even though this is Draft, the test does work as-is and the waivers should be mostly accurate.

I didn't apparently need to use any skip-tags, the runs finished in a reasonable amount of time and there were no failures that I could easily identify as part of the testing environment.  
There were a few spurious "could not resolve" DNS issues with our internal download servers, but those seem intermittent and didn't always manifest.

If you know of something that should definitely be in skip-tags, let me know, otherwise I'll remove the TODO along with the commented-out tags. The functionality will remain, just with empty list for now.

The `same_os` skip-tags are useless for now (they're for a future "machine hardening" style test), but since I went through the old tests, I put them here anyway.

Otherwise the Ansible test really just does `ansible-playbook` + check via `oscap` (like for Anaconda).

Your comments on the waivers are welcome, but, as usual, I'm not aiming to investigate every single fail here in this PR. It's mostly about implementing a new test + recording what currently fails.

Any extra non-Ansible changes are confined to their commits.

Thanks!

### DIFF
--- a/hardening/ansible/test.py
+++ b/hardening/ansible/test.py
@@ -27,15 +27,23 @@ if not g.can_be_snapshotted():
     g.install(kickstart=ks)
     g.prepare_for_snapshot()
 
-# TODO: no clue why this is needed, but it was in the old tests
-for collection in ['community.general', 'ansible.posix']:
-    util.subprocess_run(
-        ['ansible-galaxy', '-vvv', 'collection', 'install', collection],
-        check=True)
-
 # the VM guest ssh code doesn't use $HOME/.known_hosts, so Ansible blocks
 # on trying to accept its ssh key - tell it to ignore this
 os.environ['ANSIBLE_HOST_KEY_CHECKING'] = 'False'
+
+# rhc-worker-playbook modules, exported per official instructions on
+# https://access.redhat.com/articles/remediation
+os.environ['ANSIBLE_COLLECTIONS_PATH'] = \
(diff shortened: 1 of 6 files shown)


hardening/ansible/test.py
```diff
@@ -0,0 +1,69 @@
...
+    g = virt.Guest('minimal_with_oscap')
+
+if not g.can_be_snapshotted():
+    ks = virt.Kickstart()
+    if use_gui:
+        ks.add_package_group('Server with GUI')
+    g.install(kickstart=ks)
+    g.prepare_for_snapshot()
+
+# TODO: no clue why this is needed, but it was in the old tests
```
matusmarhefka, line 30: We have a function which documents the proper way how to install Ansible + additional modules: https://src.fedoraproject.org/tests/scap-security-guide/blob/main/f/Library/scap-common/lib.sh#_202

Basically, `ansible-core` in RHEL/CS doesn't contain all the modules which are used in the scap-security-guide  so we need to either install `rhc-worker-playbook` RPM (RHEL) or install those modules using `ansible-galaxy` (CentOS Stream).

comps: I switched the code to just use `rhc-worker-playbook` alone, per https://access.redhat.com/articles/remediation. If/when we need to run on CentOS Stream, we can easily uncomment it and add a condition.

matusmarhefka (approved):

PR #20, opened 2023-06-11
Add `/scanning/oscap-eval` as a basic smoke test

comps:
There are a few more commits included, notably

* `split lib.util to several sub-modules, for clarity`
  * I originally did this because I was making more XML/HTML parsing functions, assuming this would be openscap-validation testing, but since that was (for now) dropped, I didn't add any functions. Still, splitting `util.py` was probably not a bad idea.
  * I see the difference between `lib/*.py` and `lib/util/*.py` as the former being bigger library categories, potentially with many functions, and the latter being per-functionality files, with only minimal supporting functions (if any). ... I wanted to avoid having 20+ files directly in `lib/` and a test developer not knowing which ones are more important for writing a test.
* `don't use more vcpus for a VM than available`
  * Just a small optimization if running on a VM with just 1 CPU, to not spawn a nested VM with 2 CPUs.
* `allow running virtual guests on 4GiB RAM hosts`
  * This is mostly for compatibility with other (not ours) Beaker HW. It's not a good idea for us to downsize VMs from 5GB to 4GB because of the increased SSD writes, but many more Beaker systems have just 4GB, so this improves compatibility.

(See commit messages for more.)

The rest is directly related to the `/scanning/scap-eval` test itself.

In addition to running a scan, the test reports each found `W` or `WARNING` or `E` or `ERROR` as a separate result, with `WARNING`s being translated to TMT `warn` (not needing a waiver), and `ERROR`s being `fail`, needing a waiver. The full text is provided as a `note`.

Ie.
```
warn /scanning/oscap-eval/WARNING (on default-0) (WARNING: Datastream component 'scap_org.open-scap_cref_security-data-oval-com.redhat.rhsa-RHEL7.xml.bz2' points out to the remote 'https://access.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2'. Use '--fetch-remote-resources' option to download it.) [1/1]
warn /scanning/oscap-eval/WARNING (on default-0) (WARNING: Skipping 'https://access.redhat.com/security/data/oval/com.redhat.rhsa-RHEL7.xml.bz2' file which is referenced from datastream) [1/1]
warn /scanning/oscap-eval/WARNING (on default-0) (WARNING: Skipping ./security-data-oval-com.redhat.rhsa-RHEL7.xml.bz2 file which is referenced from XCCDF content) [1/1]
warn /scanning/oscap-eval/WARNING (on default-0) (W: oscap: Obtrusive data from probe!) [1/1]
warn /scanning/oscap-eval/WARNING (on default-0) (W: probe_environmentvariable58: Entity has no value!) [1/1]
```

See also the new `conf/waivers` addition, which I don't think was found before due to how broken `oscap` is on RHEL-7 (see code comments).

The thing is -- is this kind of reporting fine, or should we treat all `WARNING`s as `fail` and explicitly waive them too? ... As far as I can tell, the old test does it like that ... ?


mildas:
> The thing is -- is this kind of reporting fine, or should we treat all WARNINGs as fail and explicitly waive them too? ... As far as I can tell, the old test does it like that ... ?

Treating warnings as `warn` is fine. LGTM so merging

PR #21, opened 2023-06-19
Add `/hardening/host-os/{oscap,ansible}` tests

comps:
This test does remediation + scan on the "host OS", the operating system running the test itself. Basically equivalent to the old "machine-hardening" test.

Supporting code (reboot, etc.) is in separate commits.

I'm not super happy with the random failures, and there are probably more that will pop up, but until we have a better way to investigate them (switch to parsing results from results ARF XML, which should also contain remediation shell script output).

### DIFF
--- a/hardening/host-os/ansible/main.fmf
+++ b/hardening/host-os/ansible/main.fmf
@@ -1,0 +1,110 @@
+summary: Runs ansible remediation directly on the target system
+test: python3 -m lib.runtest ./test.py
+result: custom
+environment+:
+    PYTHONPATH: ../../..
+duration: 1h
+recommend+:
+    # ansible-core is not available on RHEL-7
+    # ansible is not on RHEL-8+
+    - ansible-core
+    - ansible
+    # needed for the ini_file ansible plugin, and more
+    - rhc-worker-playbook
+enabled: false
+
+/anssi_bp28_high:
+    environment+:
+        PROFILE: anssi_bp28_high
+    adjust:
+        when: distro == rhel-7
+        enabled: false
+        because: RHEL-7 has anssi_nt28_high instead
+    extra-summary: /CoreOS/scap-security-guide/hardening/host-os/ansible/anssi_bp28_high
+
+/anssi_nt28_high:
+    environment+:
+        PROFILE: anssi_nt28_high
+    adjust:
+        when: distro >= rhel-8
+        enabled: false
+        because: RHEL-8 and newer have anssi_bp28_high instead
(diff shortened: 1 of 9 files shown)


comps:
Switching to draft, I'll add the ansible version here too, since it wasn't that difficult to make.

comps:
The ansible test is disabled for now due to how many un-investigated failures we already have for the VM-based ansible test, https://github.com/RHSecurityCompliance/contest/blob/428bd1abe8f3863b94aea197a111e4f3563b39c2/conf/waivers#L101-L164 and many (all?) of the `host-os/ansible` failures seem to be the same as for this other test.

I will file a Github issue (after this is merged) for investigating the VM-based ansible failures, and include enabling `host-os/ansible` as part of the description.

hardening/host-os/oscap/test.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # restore basic login functionality
+    util.subprocess_run(['chage', '-d', '99999', 'root'], check=True)
+    with open('/etc/sysconfig/sshd', 'a') as f:
+        f.write('\nOPTIONS=-oPermitRootLogin=yes\n')
+
+    util.reboot()
+
+else:
+
```
mildas, line 41: Please remove empty lines around `else` and after `if`

hardening/host-os/ansible/main.fmf
```diff
@@ -0,0 +1,110 @@
...
+environment+:
+    PYTHONPATH: ../../..
+duration: 1h
+recommend+:
+    # ansible-core is not available on RHEL-7
+    # ansible is not on RHEL-8+
+    - ansible-core
+    - ansible
+    # needed for the ini_file ansible plugin, and more
+    - rhc-worker-playbook
```
mildas, line 13: We should agree on `recommend`/`required` indentation. `contest/main.fmf` has no indentation and so ctest proposed by me. But `hardening/*/main.fmf` has 4-space.

comps: I believe I was using TMT's version of indenting. When you actually export to TCMS (by letting TMT rewrite your `.fmf` file), it inserts this kind of 4-space indent into lists.

As far as I can tell, this is mostly against official YAML code style, which uses ie.
```yaml
recommend+:
- ansible-core
- ansible
- ...
```
instead of
```yaml
recommend+:
    - ansible-core
    - ansible
    - ...
```
But I believe my version is more TMT-friendly (would need to recheck).

mildas: I'm fine with that. I'll update ctest PR with the 4-space indentation

mildas (approved):

PR #22, opened 2023-06-20
Add /static-checks/ansible/allowed-modules

matusmarhefka:
New test migrated from
https://src.fedoraproject.org/tests/scap-security-guide/blob/main/f/Sanity/ansible-allowed-modules


static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+
+    with open(datastream) as ds:
+        with tempfile.NamedTemporaryFile(mode='w', suffix='.xml') as new_ds:
+            for line in ds:
+                if any(x.search(line) for x in exprs):
+                    line = line.replace('selected="false"', 'selected="true"')
+                new_ds.write(line)
+            new_ds.flush()
+            util.log(f"Saving modified {datastream} with all rules being selected as"
+                     "{new_ds.name}")
```
comps, line 46: Our code style permits <= 99 characters on a line, you don't need to break the string here.
```python
util.log(f"Saving modified {datastream} with all rules being selected as {new_ds.name}")
```
While we try to keep docstrings <= 80 (as these are often seen on a console window via `pydoc`), most of contest code actually heavily uses >80 lines.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+            yield Path(new_ds.name)
+
+
+def list_append_ansible_module_from_task(modules_list, task_keywords, allowed_modules):
```
comps, line 50: Are you sure you want this as a separate function?

It's called only from `traverse_blocks()` (where it's contained in one `else:` block), it gets all its arguments, etc. .. It could be better to just include the function body in `traverse_blocks()`.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    with open(playbook_path, 'r') as yamlfile:
+        playbook = yaml.safe_load(yamlfile)
+    modules_list = []
+
+    util.log(f"Check Ansible playbook {playbook_path} contains only allowed modules")
+    for section in playbook:
+        if section['tasks'] is None:
+            # Playbook has no tasks.
+            return
```
comps, line 101: This aborts processing of other sections in the playbook. Maybe `continue` would be better?

If there is always only one section in every playbook (?), then don't use a `for` loop, a simple `section = next(playbook)` or `section = next(iter(playbook))` would more explicitly state there is only one section.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+        for task in section['tasks']:
+            traverse_blocks(modules_list, task, allowed_modules)
+
+    modules_list.sort()
+    for m in modules_list:
+        results.report('pass', m, 'module is allowed')
```
comps, line 107:
```python
for m in sorted(modules_list):
```
would probably be faster/cleaner given that the sorted `modules_list` is not reused elsewhere.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+# https://access.redhat.com/articles/remediation
+os.environ['ANSIBLE_COLLECTIONS_PATH'] = \
+    '/usr/share/rhc-worker-playbook/ansible/collections/ansible_collections/'
+
+with select_all_rules(util.get_datastream()) as ds_file:
+    with open(ds_file) as ds:
+        with tempfile.NamedTemporaryFile(mode='w', suffix='.yml') as playbook:
+            util.log(f"Generate Ansible playbook {playbook.name} for all rules from {ds_file}")
+            oscap_cmd = ['oscap', 'xccdf', 'generate', 'fix', '--fix-type', 'ansible', '--output',
+                         playbook.name, ds.name]
```
comps, line 120: In contest, we generally try to avoid this style of line breaking as it leads to excessive indent, and is discouraged by PEP8 (or its extensions?).

Consider this instead:
```python
oscap_cmd = [
    'oscap', 'xccdf', 'generate', 'fix', '--fix-type', 'ansible',
    '--output', playbook.name,
    ds.name,
]
```

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+            util.subprocess_run(oscap_cmd, check=True)
+
+            ansible_cmd = ['ansible-playbook', '--syntax-check', playbook.name]
+            util.subprocess_run(ansible_cmd, check=True)
```
comps, line 124: A minor thing, but I would rather include the `[]` inside `util.subprocess_run()` if it fits on a single line.

In this case, it would be exactly 98 characters, which is almost the maximum allowed, but still allowed. :smile:
```python
            util.subprocess_run(['ansible-playbook', '--syntax-check', playbook.name], check=True)
```

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+
+            # Obtains a set allowed_modules containing all the module names provided by RPM
+            # package(s) and/or Ansible collections which will be used as a list of allowed
+            # modules.
+            _, allowed_modules = util.subprocess_stream('./get_all_ansible_modules.sh',
+                                                        stderr=subprocess.STDOUT, check=True)
```
comps, line 130: I would use the PEP8 recommended function argument multiline syntax to avoid the extra indent here,
```python
_, allowed_modules = util.subprocess_stream(
    './get_all_ansible_modules.sh', stderr=subprocess.STDOUT, check=True)
```
or
```python
_, allowed_modules = util.subprocess_stream(
    './get_all_ansible_modules.sh',
    stderr=subprocess.STDOUT,
    check=True)
```
(I don't like this style too much, but it's a python standard.)

Additionally - do you really want to redirect `stderr` to `stdout`? The script doesn't seem to produce anything on `stderr` and leaving it on the console (TMT log) by removing the `stderr=` is probably safer and more useful.  
It would also make the line 99 characters, so it would fit on one line. :smile:

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+def check_allowed_modules(playbook_path, allowed_modules):
+    """
+    Collects all modules from Ansible playbook playbook_path and reports pass
+    if a module is found in the allowed_modules set, otherwise reports fail
+    and prints an offending Ansible module and task. Function also handles
+    Ansible block sections which can bundle multiple tasks.
+    Note: ansible playbook must be a valid playbook which means that
+    'ansible-playbook --syntax-check' must not report any errors for it.
+    """
+    with open(playbook_path, 'r') as yamlfile:
```
comps, line 93: This can be just
```python
with open(playbook_path) as yamlfile:
```
The `r` is default and typically omitted.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+        playbook = yaml.safe_load(yamlfile)
+    modules_list = []
+
+    util.log(f"Check Ansible playbook {playbook_path} contains only allowed modules")
+    for section in playbook:
+        if section['tasks'] is None:
+            # Playbook has no tasks.
+            return
+        for task in section['tasks']:
+            traverse_blocks(modules_list, task, allowed_modules)
```
comps, line 103: Please don't use the C-style pointer passing of `modules_list`, filling it in nested functions.

The pythonic style is to atomically return a portion to be appended (one or more entries, as a generator or a list) and append it here in the parent function, ie.
```python
for task in section['tasks']:
    modules_list += traverse_blocks(task, allowed_modules)
```

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+# Obtained from
+# https://docs.ansible.com/ansible/latest/reference_appendices/playbooks_keywords.html#task
+ansible_reserved_keywords = ['action', 'any_errors_fatal', 'args', 'async', 'become', 'become_exe',
+                             'become_flags', 'become_method', 'become_user', 'changed_when',
+                             'check_mode', 'collections', 'connection', 'debugger', 'delay',
+                             'delegate_facts', 'delegate_to', 'diff', 'environment', 'failed_when',
+                             'ignore_errors', 'ignore_unreachable', 'local_action', 'loop',
+                             'loop_control', 'module_defaults', 'name', 'no_log', 'notify', 'poll',
+                             'port', 'register', 'remote_user', 'retries', 'run_once', 'tags',
+                             'throttle', 'timeout', 'until', 'vars', 'when', 'with_items']
```
comps, line 23: Would it make sense to reformat this into something more manageable?
```python
ansible_reserved_keywords = [
    'action',
    'any_errors_fatal',
    'args',
    'async',
    ...
]
```
or at least avoid the indent, and use breaking at 99-char lines?
```python
ansible_reserved_keywords = [
    'action', 'any_errors_fatal', 'args', 'async', 'become', 'become_exe', 'become_flags',
    'become_method', 'become_user', 'changed_when', 'check_mode', 'collections', 'connection',
    ...
]
```
?

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,134 @@
...
+    return 0
+
+
+def traverse_blocks(modules_list, task, allowed_modules):
+    block_set = set(['block'])
+    task_keywords = set(task.keys())
+
+    if len(task_keywords.intersection(block_set)) > 0:
+        for block_task in task['block']:
+            return traverse_blocks(modules_list, block_task, allowed_modules)
```
comps, line 75: Same as for traversing tasks in the parent function, `return` here (and in the `else` block) doesn't make sense with the `for`. Maybe you meant to `yield` blocks to parent or not `return` at all?

comps (changes requested):
There were also a few `"` vs `'` comment inconsistencies that I didn't mention because it would clutter the review, but you might want to do a brief read-through and see whether `"` vs `'` match the code style (`"` for human text, `'` for identifiers).

I also didn't comment on everything and I have no clue how the playbooks are structured, so this is mostly a surface-level review. Hope it helps anyway.

I can do a second review round based on what you change.

static-checks/ansible/allowed-modules/test.py
matusmarhefka: Yeah, good catch!

static-checks/ansible/allowed-modules/test.py
matusmarhefka: I changed it a bit, function now returns either a string or None.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,127 @@
...
+            yield Path(new_ds.name)
+
+
+def traverse_blocks(task, allowed_modules):
+    block_set = set(['block'])
+    task_keywords = set(task.keys())
+
+    if len(task_keywords.intersection(block_set)) > 0:
+        for block_task in task['block']:
+            traverse_blocks(block_task, allowed_modules)
```
comps, line 54: This should probably be
```python
return traverse_blocks(block_task, allowed_modules)
```
otherwise you lose all info about nested blocks/tasks.

comps (changes requested):

static-checks/ansible/allowed-modules/test.py
matusmarhefka: Hmm, but it was your suggestion to remove it :D https://github.com/RHSecurityCompliance/contest/pull/22#discussion_r1235483926

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,127 @@
...
+        task_module = list(updated_keywords.intersection(allowed_modules))
+        if len(task_module) == 1:
+            return str(task_module[0])
+        else:
+            module = ''.join(
+                [x for x in list(updated_keywords) if x not in ansible_reserved_keywords]
+            )
+            results.report('fail', module, 'module is not allowed')
+            util.log(f"This occurred in the following task:\n{task}\n")
+            return None
```
mildas, line 69: This part is confusing.

Comparing it to https://src.fedoraproject.org/tests/scap-security-guide/blob/main/f/Sanity/ansible-allowed-modules/check_allowed_modules.py#_6 there is nothing about reserved keywords. Why is it needed there? And why is it used to filter our keywords right before reporting `fail`? What if `task_module` has multiple keywords, one of them is allowed module and others are reserved keywords? In that case this false positive.

static-checks/ansible/allowed-modules/get_all_ansible_modules.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
```
mildas, line 2: I believe this could be rewritten to Python3. Thus avoiding Bash script there and making it even nicer - getting rid of multiple pipes that are thee times there:
```
for package in ['ansible', 'ansible-core', 'rhc-worker-playbook']:
    ...
```
similarly for `else` part.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,127 @@
...
+                modules_list.append(module)
+
+    for m in sorted(modules_list):
+        results.report('pass', m, 'module is allowed')
+
+
+# rhc-worker-playbook modules, exported per official instructions on
+# https://access.redhat.com/articles/remediation
+os.environ['ANSIBLE_COLLECTIONS_PATH'] = \
+    '/usr/share/rhc-worker-playbook/ansible/collections/ansible_collections/'
```
mildas, line 103: This is RHEL-specific. For Fedora/CentOS we need https://github.com/RHSecurityCompliance/contest/blob/main/hardening/host-os/ansible/test.py#L26 Probably as lib because it will be needed on more places.

Not a big deal now, but mentioning because of inconsistency - `get_all_ansible_modules.sh` is ready for Fedora/CentOS and this part not.

static-checks/ansible/allowed-modules/test.py
comps: Okay, good point. This time around, I focused on recursion and didn't see the `for`. :smile: 

The last sentence is however correct. The overall logic should be something like
```python

found_allowed_modules = set()

for all items in the current task:
    if item is a block:
        found_allowed_modules += traverse_blocks(...)
    else:
        modules_in_current_item = ....
        if modules_in_current_item contains forbidden words:
            report fail
        else:
            found_allowed_modules += modules_in_current_item

return found_allowed_modules
```
Basically collect allowed modules from both nested blocks as well as current task list items.

(If you want to be fancy, you don't actually need `found_allowed_modules` at all, you can just do `yield from ...`, but I wanted to illustrate what I meant.)

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,127 @@
...
+    block_set = set(['block'])
+    task_keywords = set(task.keys())
+
+    if len(task_keywords.intersection(block_set)) > 0:
+        for block_task in task['block']:
+            traverse_blocks(block_task, allowed_modules)
+    else:
+        updated_keywords = set()
+        for keyword in task_keywords:
+            keyword = keyword.replace('ansible.builtin.', '')
```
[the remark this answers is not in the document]
comps: No, `removeprefix` is not in Python 3.6 unfortunately.

static-checks/ansible/allowed-modules/test.py
comps: I think that's fine. This test is clearly heavily based on something existing, and our other ansible tests are RHEL-only too (the CentOS parts are commented out for now).

matusmarhefka: We should address this for all `contest` tests, so I will leave it as is for now.

static-checks/ansible/allowed-modules/test.py
matusmarhefka: It is there to identify name of the module which is not allowed. When you won't find any allowed module you don't know which item from the python dictionary represents a valid module name and what is just Ansible keyword (used in tasks, e.g. `when`, `tags`, etc.). Therefore, we have a list of all allowed keywords in tasks and we filter them out which will leave us with one item and that should be the name of the module.

matusmarhefka: I will add a comment with explanation.

matusmarhefka:
> Rather than add further noise into the current code with more comments, I would suggest choosing a different approach to traversing the tasks.
> 
> Split the traversal from the allowance check, don't overuse `set()` methods where not necessary, etc.
> 
> How about something like this (untested)?
>  ...

@comps I didn't use your suggestion, but I reworked the code to use sets instead of lists as they are better fit here and also addressed other comments.

static-checks/ansible/allowed-modules/get_all_ansible_modules.sh
matusmarhefka: Makes sense, I will look into it.

comps:
Okay, I think I finally understand what's going on. Sort of.

It still seems weird to me, because it heavily relies on the fact that a valid task (list of keywords) doesn't have more than 1 module and the logic completely breaks down on multiple places if that happens. But that may be okay.

----
```python
allowed_module = keywords.intersection(all_allowed_modules)
```
This counts on the fact that a set of ~ 5 keywords will have only **one** intersection result with the list of modules (which doesn't include keywords).  
If it accidentally has 2 or more modules (some of them may be forbidden), the code that follows actually adds all good module names into `found_allowed_modules` and silently ignores the forbidden ones.

----

```python
forbidden_module = keywords.difference(ansible_reserved_keywords).pop()
```
This then filters out the ~ 4 other keywords, leaving just the one module name.  
What confused me is that `ansible_reserved_keywords` is never actually checked against - it's used only for filtering out the remaining keywords, you never check whether the rest of the keywords is valid (contained in `ansible_reserved_keywords`).

I still think there's a better/simpler way to write the algorithm, but at least now I understand your version and it LGTM, given the limitations above.

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,121 @@
...
+    original_keywords = set(task.keys())
+    found_allowed_modules = set()
+
+    if 'block' in original_keywords:
+        for block_task in task['block']:
+            found_allowed_modules.update(process_task(block_task, all_allowed_modules))
+    else:
+        keywords = set()
+        for keyword in original_keywords:
+            keywords.add(keyword.replace('ansible.builtin.', ''))
```
comps, line 58: Could be perhaps compacted as
```python
keywords = set(x.replace('ansible.builtin.', '') for x in original_keywords)
```

static-checks/ansible/allowed-modules/test.py
```diff
@@ -0,0 +1,121 @@
...
+        if allowed_module:
+            found_allowed_modules.update(allowed_module)
+        else:
+            # No allowed module found in 'keywords' means there is a module
+            # which is forbidden and we obtain it by filtering out all the
+            # reserved keywords for Ansible tasks which will leave us with
+            # one item and that should be the name of that forbidden module.
+            forbidden_module = keywords.difference(ansible_reserved_keywords).pop()
+            task_name = task['name'] if 'name' in task else ''
+            results.report('fail', forbidden_module, f'forbidden module in task \'{task_name}\'')
```
comps, line 69: I would avoid reporting kind of useless `note` for the results (here and for `pass` as well). A note being present in the HTML distorts the output somewhat and is just visual clutter in this case.

The test is named `allowed-modules`, so any failures should be forbidden modules, and any passes allowed modules, no need to verbosely state that in the note.

But I'll leave the decision up to you.

matusmarhefka:
> 

As far as I know, a valid task can only specify one module, so that shouldn't be a problem. And if that would happen, test would fail on `ansible-playbook --syntax-check` which runs before checking allowed modules :)

static-checks/ansible/allowed-modules/test.py
matusmarhefka: Ok, I will update it.

mildas (approved):

PR #23, opened 2023-06-20
"smoke-test" - build from src rpm and ctest it

mildas:
Src rpm download and build is done in new `package.py` library. Dunno, if it fits well there as the test might be the only place where the functions will be used. In such case, I can move them to test script.

No new custom logging function was implemented in library, because this form of results is specific for ctest -- get logs to file using `--output-log ctest_results` ctest option, process only test result lines from the file, parse name+result and report it.

### DIFF
--- a/static-checks/rpmbuild-ctest/test.py
+++ b/static-checks/rpmbuild-ctest/test.py
@@ -1,0 +1,68 @@
+#!/usr/bin/python3
+
+import re
+from pathlib import Path
+
+from lib import util, results, versions
+
+
+def download_source(package):
+    """
+    Download src rpm and return Path of the file.
+
+    'package' is name of package to download.
+    """
+    util.subprocess_run(['yumdownloader', '--source', package], check=True)
+    package_glob = package + '*.src.rpm'
+    try:
+        src_path = next(Path.cwd().glob(package_glob))
+    except StopIteration:
+        raise FileNotFoundError(f".src.rpm not found for {package}")
+    return src_path
+
+
+def build_source(src_rpm, path=None):
+    """
+    Build package from src rpm.
+
+    'src_rpm' is path to src rpm file.
+    Define 'path' to specify path where to build. Otherwise, build
+    is done to rpmbuild directory in current working directory.
+    """
+    builddep_cmd = ['yum-builddep'] if versions.rhel == 7 else ['dnf', 'builddep']
+    util.subprocess_run([*builddep_cmd, '-y', src_rpm], check=True)
+
+    path = Path(path) if path else Path.cwd() / 'rpmbuild'
+
+    util.subprocess_run(['rpm', '-ivh', '--define', f'_topdir {path}', src_rpm], check=True)
+
+    spec_path = next((path / 'SPECS').glob('*.spec'))
+    util.subprocess_run(['rpmbuild', '-bc', '--define', f'_topdir {path}', spec_path], check=True)
+
+    return path
+
+
+# Get src rpm and build from it
+src_rpm = download_source('scap-security-guide')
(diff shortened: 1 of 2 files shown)


comps:
First, flake8:
```
./static-checks/rpmbuild-ctest/test.py:15:70: F541 f-string is missing placeholders
./static-checks/rpmbuild-ctest/test.py:15:100: E501 line too long (124 > 99 characters)
```

As far as the separate library - I think I would prefer to have it bundled with the test.  
If you want it in a separate file (and not clutter test code), you should be able to just move `package.py` to the test directory and import it as `import package` (or `from . import package`?) since python3 implicitly adds `.` into `sys.path`.

I'll add code comments inline.

lib/util/package.py
```diff
@@ -0,0 +1,50 @@
...
+
+    'package' is name of package to download.
+    """
+    cmd = ['yumdownloader', '--source', package]
+    subprocess_run(cmd, check=True)
+    package_glob = package + "*.src.rpm"
+    try:
+        src_path = next(Path('.').glob(package_glob))
+    except StopIteration:
+        return None
```
comps, line 20: Actually, I would throw a `FileNotFoundError` here. The test code doesn't check for `None` and we probably want to abort if this operation fails anyway.

lib/util/package.py
```diff
@@ -0,0 +1,50 @@
...
+    if versions.rhel == 7:
+        yum_builddep = ['yum-builddep']
+    else:
+        yum_builddep = ['dnf', 'builddep']
+
+    cmd = [*yum_builddep, '-y', src_rpm]
+    subprocess_run(cmd, check=True)
+
+    if not path:
+        path = os.getcwd() + "/rpmbuild"
```
comps, line 41: Given that our current "policy" is to use `pathlib` where possible, this would be more preferable:
```python
path = Path(path) if path else Path.cwd() / 'rpmbuild'
```

static-checks/rpmbuild-ctest/test.py
```diff
@@ -0,0 +1,31 @@
...
+from lib import util, results
+
+
+# Get src rpm and build from it
+src_rpm = util.download_source('scap-security-guide')
+rpm_build = util.build_source(src_rpm)
+ssg_build = str(next(Path(f'{rpm_build}/BUILD').glob('*/build')))
+
+# ctest
+cmd = ['cmake', '--build', f'{ssg_build}', '--target', 'test', '--', f'ARGS=--output-on-failure --output-log ctest_results']
```
comps, line 15: In addition to the flake8 reports, `f'{ssg_build}` should be just `ssg_build`.

comps (changes requested):
I didn't comment on every single `"` vs `'`, but you might want to do a quick read-through to see whether they meet our style rules for human-language vs identifier, as I saw a few outliers.

static-checks/rpmbuild-ctest/test.py
```diff
@@ -0,0 +1,68 @@
...
+    Download src rpm and return Path of the file.
+
+    'package' is name of package to download.
+    """
+    util.subprocess_run(['yumdownloader', '--source', package], check=True)
+    package_glob = package + '*.src.rpm'
+    try:
+        src_path = next(Path.cwd().glob(package_glob))
+    except StopIteration:
+        return FileNotFoundError
```
comps, line 20: I don't think this is going to work. :smile:  
Try:
```python
    except StopIteration:
        raise FileNotFoundError(f".src.rpm not found for {package}")
```

mildas: Oh god :D Right

PR #26, opened 2023-06-25
Update waives, prepare for real-world use

comps:
This hopefully prepares contest for use, replacing Beakerlib-based tests.

The waivers were updated to match current upstream state (2 commits), and given enough re-runs to filter out random openscap crashes/freezes, they should return no failures. Due to hardware limitations, `/hardening/host-os` wasn't as rigorously tested.

Then there's the waiver file split - while we don't have a perfect process/idea defined for how/when to move `upstream` to `released`, we can likely deal with that while defining (updating?) a new downstreaming process. I feel the split makes the state more predictable, as changes to `waivers-upstream` have no chance to accidentally affect `waivers-released`, which is a risk when both stay in the same file. Amongst other things.

Finally, all the new tests have been exported to TCMS/Nitrate and the FMF files carefully hand-edited to keep comments, empty lines, and proper line breaks for `because:`.  
The indent change comes from TMT, so I left it in. No clue if that's a recent TMT change (shifting lists 2 spaces left, to keep the visual indent at 4 spaces), possibly so.

### DIFF
--- a/conf/waivers
+++ b/conf/waivers
@@ -74,10 +74,6 @@
 /hardening/oscap/[^/]+/package_rear_installed
     Match(rhel <= 8, sometimes=True)
 
-# https://github.com/ComplianceAsCode/content/issues/10592
-/hardening/anaconda/with-gui/cis_workstation_l[12]/file_groupownership_sshd_private_key
-    ssg >= '0.1.68'
-
 # RHEL-8: https://bugzilla.redhat.com/show_bug.cgi?id=1834716
 # RHEL-9: https://bugzilla.redhat.com/show_bug.cgi?id=1999587
 /hardening/anaconda/with-gui/[^/]+/service_avahi-daemon_disabled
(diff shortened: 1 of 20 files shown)


conf/waivers
```diff
@@ -182,6 +182,6 @@
 
 # TODO: is this a reported bug?
 /scanning/oscap-eval/ERROR
-    rhel == 7 and note == 'E: oscap: Failed to convert OVAL state to SEXP, id: oval:ssg-state_file_groupowner_var_log_syslog_gid_4_0:ste:1.'
+    rhel <= 8 and note == 'E: oscap: Failed to convert OVAL state to SEXP, id: oval:ssg-state_file_groupowner_var_log_syslog_gid_4_0:ste:1.'
```
mildas, line 174: https://bugzilla.redhat.com/show_bug.cgi?id=1797653 We can re-open it.

mildas (approved):

PR #31, opened 2023-06-29
Update waivers for beaker, from TCMS/wow

comps:
Aside from some small fixes, this is mostly updating waivers to account for Beaker-scheduled tests and `/hardening/host-os` running on actual Beaker-based systems, instead of custom-provisioned virtual machines.

It's not pretty, and a lot of it had to be made `sometimes=True`, but I couldn't get it stabilized otherwise. The fails were just too random.

We'll have to deal with them as we investigate all the TODOs in waivers.

### DIFF
--- a/conf/waivers-released
+++ b/conf/waivers-released
@@ -9,15 +9,19 @@
 /hardening/oscap(/with-gui)?/[^/]+/configure_bashrc_exec_tmux
 /hardening/oscap(/with-gui)?/[^/]+/no_tmux_in_shells
 /hardening/oscap(/with-gui)?/[^/]+/configure_usbguard_auditbackend
-/hardening/host-os/oscap/[^/]+/configure_bashrc_exec_tmux
-/hardening/host-os/oscap/[^/]+/no_tmux_in_shells
-/hardening/host-os/oscap/[^/]+/configure_usbguard_auditbackend
     rhel >= 8
 /hardening/oscap(/with-gui)?/[^/]+/configure_bashrc_tmux
 /hardening/oscap(/with-gui)?/stig(_gui)?/postfix_prevent_unrestricted_relay
+    rhel == 8
+
+# same issue, but host-os seems to be a lot more random in this
+/hardening/host-os/oscap/[^/]+/configure_bashrc_exec_tmux
+/hardening/host-os/oscap/[^/]+/no_tmux_in_shells
+/hardening/host-os/oscap/[^/]+/configure_usbguard_auditbackend
+    Match(rhel >= 8, sometimes=True)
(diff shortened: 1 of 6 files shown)


hardening/host-os/oscap/test.py
```diff
@@ -15,15 +16,23 @@
 ds = util.get_datastream()
 
 unique_name = util.get_test_name().lstrip('/').replace('/', '-')
-new_ds = Path(f'/var/tmp/contest-{unique_name}')
+
```
mildas, line 19: The empty line is not needed there

hardening/oscap/test.py
```diff
@@ -29,7 +29,8 @@
     g.copy_to(util.get_datastream(), 'contest-ds.xml')
 
     # remediate, reboot
-    g.ssh(f'oscap xccdf eval --profile {profile} --progress --remediate contest-ds.xml')
+    g.ssh(f'oscap xccdf eval --profile {profile} --progress '
+          f'--report remediation.html --remediate contest-ds.xml')
```
mildas, line 33: No need for f-string, right?

conf/waivers-released
```diff
@@ -72,29 +76,60 @@
...
-# probably because Beaker has crond enabled
-# - TODO: shouldn't oscap be able to disable a service via remediation?
-/hardening/host-os/oscap/[^/]+/service_crond_enabled
+# Beaker-specific:
+# all Beaker repositories have gpgcheck=0 and they get copied to nested VMs too
+/(hardening|scanning)/.+/ensure_gpgcheck_never_disabled
+# we don't control partitions on the host OS
+/hardening/host-os/oscap/.+/mount_option_(home|opt|srv|var|var_log|var_log_audit)_(noexec|nosuid|nodev|usrquota|grpquota)
+# likely something caused by restraint / Beaker test env 
+/hardening/host-os/oscap/.+/file_permissions_unauthorized_world_writable
```
mildas, line 103: You can add there https://github.com/beakerlib/beakerlib/issues/149

hardening/host-os/oscap/test.py
comps: It's not very visible from the quoted code, but the context is:
```python
unique_name = util.get_test_name().lstrip('/').replace('/', '-')

# persistent across reboots
tmpdir = Path(f'/var/tmp/contest-{unique_name}')

new_ds = tmpdir / 'modified_datastream.xml'
```
I added the newline to prevent somebody associating the "persistent across reboots" comment with the `new_ds` line. But I guess I can move it closer, the comment wouldn't be wrong per se.

PR #46, opened 2023-08-01
For RHEL7 generate also XCCDF results

mildas:
Old openscap in RHEL7 doesn't put OVAL findings to HTML report. To make investigation easier there, include XCCDF results from `--results`.

### DIFF
--- a/hardening/host-os/ansible/test.py
+++ b/hardening/host-os/ansible/test.py
@@ -54,11 +54,13 @@ else:
     # old RHEL-7 oscap mixes errors into --progress rule names without a newline
     verbose = ['--verbose', 'INFO'] if versions.oscap >= 1.3 else []
     redir = {'stderr': subprocess.STDOUT} if versions.oscap >= 1.3 else {}
+    # RHEL-7 HTML report doesn't contain OVAL findings by default
+    oval_results = [] if versions.oscap >= 1.3 else ['--results', 'results.xml', '--oval-results']
 
     # scan the remediated system
     cmd = [
         'oscap', 'xccdf', 'eval', *verbose, '--profile', profile,
-        '--progress', '--report', 'report.html',
+        '--progress', '--report', 'report.html', *oval_results,
         util.get_datastream(),
     ]
     proc, lines = util.subprocess_stream(cmd, **redir)
--- a/hardening/host-os/oscap/test.py
+++ b/hardening/host-os/oscap/test.py
@@ -51,11 +51,13 @@ else:
     # old RHEL-7 oscap mixes errors into --progress rule names without a newline
     verbose = ['--verbose', 'INFO'] if versions.oscap >= 1.3 else []
     redir = {'stderr': subprocess.STDOUT} if versions.oscap >= 1.3 else {}
+    # RHEL-7 HTML report doesn't contain OVAL findings by default
+    oval_results = [] if versions.oscap >= 1.3 else ['--results', 'results.xml', '--oval-results']
 
     # scan the remediated system
     cmd = [
         'oscap', 'xccdf', 'eval', *verbose, '--profile', profile,
-        '--progress', '--report', 'report.html',
+        '--progress', '--report', 'report.html', *oval_results,
         new_ds,
     ]
     proc, lines = util.subprocess_stream(cmd, **redir)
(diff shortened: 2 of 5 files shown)


hardening/host-os/ansible/test.py
```diff
@@ -54,16 +54,19 @@
     # old RHEL-7 oscap mixes errors into --progress rule names without a newline
     verbose = ['--verbose', 'INFO'] if versions.oscap >= 1.3 else []
     redir = {'stderr': subprocess.STDOUT} if versions.oscap >= 1.3 else {}
+    # RHEL-7 HTML report doesn't contain OVAL findings
+    xccdf_results = '' if versions.oscap >= 1.3 else '--results results.xml'
```
comps, line 58: This won't work with `*` expansion below. Use
```python
    xccdf_results = [] if versions.oscap >= 1.3 else ['--results', 'results.xml']
```

hardening/host-os/oscap/test.py
```diff
@@ -51,16 +51,20 @@
...
+        '--progress', '--report', 'report.html', *xccdf_results,
         new_ds,
     ]
     proc, lines = util.subprocess_stream(cmd, **redir)
     oscap.report_from_verbose(lines)
     if proc.returncode not in [0,2]:
         raise RuntimeError("post-reboot oscap failed unexpectedly")
 
+    reports = ['report.html', tmpdir / 'remediation.html'] if versions.oscap >= 1.3 \
+        else ['report.html', tmpdir / 'remediation.html', 'results.xml']
```
comps, line 69: For these longer lines, I'd personally prefer the more traditional syntax. .. Also the `>=` condition is inverted - we **want** the `results.xml` if `>= 1.3` and don't want it otherwise.
```python
if versions.oscap >= 1.3:
    logs = ['report.html', tmpdir / 'remediation.html', 'results.xml']
else:
    logs = ['report.html', tmpdir / 'remediation.html']
```
Or even
```python
logs = ['report.html', tmpdir / 'remediation.html']
if versions.oscap >= 1.3:
    logs.append('results.xml')
```

hardening/host-os/oscap/test.py
```diff
@@ -51,16 +51,20 @@
...
     results.report_and_exit(logs=['report.html', tmpdir / 'remediation.html'])
```
comps, line 70: And the above will be ignored, because this hardcodes it.

hardening/oscap/test.py
```diff
@@ -36,15 +36,19 @@
...
     oscap.report_from_verbose(lines)
     if proc.returncode not in [0,2]:
         raise RuntimeError("post-reboot oscap failed unexpectedly")
 
     g.copy_from('report.html')
     g.copy_from('remediation.html')
 
-results.report_and_exit(logs=['report.html', 'remediation.html'])
+reports = ['report.html', 'remediation.html'] if versions.oscap >= 1.3 \
+    else ['report.html', 'remediation.html', 'results.xml']
```
comps, line 53: Same as above.

mildas:
@comps Thanks for review. I've reworked it as @matusmarhefka suggested -- when `--results` is combined with `--oval-results`, it includes OVAL details to HTML report. And that's believe the best solution, so we don't need to upload huge XML file.

hardening/host-os/ansible/test.py
```diff
@@ -54,11 +54,13 @@
     # old RHEL-7 oscap mixes errors into --progress rule names without a newline
     verbose = ['--verbose', 'INFO'] if versions.oscap >= 1.3 else []
     redir = {'stderr': subprocess.STDOUT} if versions.oscap >= 1.3 else {}
+    # RHEL-7 HTML report doesn't contain OVAL findings by default
+    oval_results = '' if versions.oscap >= 1.3 else ['--results', 'results.xml', '--oval-results']
```
comps, line 58: You used only half of my correction. :smile:, but `*` expansion of an empty string seems to be somehow magically working, though using `[]` would keep the same data type.

mildas: D'oh. Fixed

comps:
Looks good now, merge when ready.

mildas:
Ok, merging. I've tested Anaconda and host-os/oscap and OVAL details are there.

PR #49, opened 2023-08-02
update waivers after first usage in productization of stabilization

vojtapolasek:
see commits for explanations.

### DIFF
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -3,6 +3,11 @@
 # - see https://github.com/ComplianceAsCode/content/issues/9746
 /hardening/anaconda(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
     rhel >= 8
+# similar case for RHEL 9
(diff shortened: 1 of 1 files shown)


comps:
Everything seems to be fine except
```
 /hardening/[^/]+/with-gui/cis_workstation_[^/]+/sysctl_net_ipv4_ip_forward
-    rhel == 8                                                                                                                                             
+    rhel <= 8    
```
which caused errors on RHEL-7 for me, indicating these to be PASS.

I wonder what changed in recent content versions to make this issue so unreliable that it fails on RHEL-7 too, needing a waiver. But either way, if you saw a genuine failure, it should probably be
```
 /hardening/[^/]+/with-gui/cis_workstation_[^/]+/sysctl_net_ipv4_ip_forward
    Match(rhel <= 8, sometimes=True)
```
to avoid the match-on-pass on RHEL-7.

comps:
Manually squashed, rebased, added comments directly to the waiver file, ... merging.

PR #52, opened 2023-08-04
decouple Beaker and FMF wrapper, support pure restraint runs

comps:
This gets rid of the concept of "running in Beaker" or "running in TMT" (mostly), replacing it with API-specific checks, **decoupling** results reporting and test name retrieval.

The reason is that multiple infrastructure pieces use the FMF wrapper (Beaker, Beaker-less restraint, etc.), so we can in fact end up with:

1. Beaker results API (LAB_CONTROLLER), and FMF wrapper
1. no Beaker results API, and FMF wrapper (standalone restraint)
1. TMT results API, no FMF wrapper
 
So I modified the reporting code to just check if it can access Beaker results API / TMT result API, and fall back to plain text.

Independently of this, test name is retrieved via TMT or FMF wrapper, no matter whether it runs under Beaker or not.

### DIFF
--- a/lib/results.py
+++ b/lib/results.py
@@ -57,6 +57,11 @@ def _sanitize_yaml_id(string):
     return re.sub(r'[^\w/ _-]', '', string, flags=re.A).strip()
 
 
+def have_tmt_api():
+    """Return True if we can report results via TMT natively."""
+    return bool(os.environ.get('TMT_TEST_DATA'))
+
+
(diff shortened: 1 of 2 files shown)


comps:
Added a second commit to clean up plain text reporting a bit, from
```
PASS /some/test / [PosixPath('/path/to/log.html'), PosixPath('/path/to/output.txt')]
```
to
```
PASS /some/test [log.html, output.txt]
```

PR #54, opened 2023-08-08
RHEL7 waiving and test skipping

mildas:
During regression run review, I've noticed `postfix_network_listening_disabled` passes all the time and the rule isn't affected by any other rule.

Also couple of all arch tests are failing on s390x/ppc64 on `ModuleNotFoundError: No module named 'urllib3'`. It's caused by missing EPEL there, so skip them.


comps:
Maybe it would be better to add this to the top-level `main.fmf` and do `adjust+` (with lists) everywhere else, so that no contest test ever runs on rhel-7 and those architectures.

But this PR is GEFN.

PR #55, opened 2023-08-11
service_rpcbind_disabled fails on rhel8 and cis_ws_l2

vojtapolasek:
https://github.com/ComplianceAsCode/content/issues/10901

### DIFF
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -210,6 +210,6 @@
 # https://github.com/ComplianceAsCode/content/issues/10901
 # not sure what enables the service, but second remediation fixes the problem
 /hardening/anaconda/with-gui/cis_workstation_l2/service_rpcbind_disabled
-    rhel == 8
+    Match(rhel == 8, sometimes=True)
 
 # vim: syntax=python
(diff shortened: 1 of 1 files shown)


conf/waivers-upstream
```diff
@@ -207,4 +207,9 @@
 /scanning/oscap-eval/ERROR
     rhel <= 8 and note == 'E: oscap: Failed to convert OVAL state to SEXP, id: oval:ssg-state_file_groupowner_var_log_syslog_gid_4_0:ste:1.'
 
+# https://github.com/ComplianceAsCode/content/issues/10901
+# not sure what enables the service, but second remediation fixes the problem
+/hardening/anaconda/with-gui/cis_workstation_l2/service_rpcbind_disabled
+    rhel == 8
```
mildas, line 213: I don't see it failing in last productization run, so it doesn't seem to be 100%, thus we need to do:
```suggestion
    Match(rhel == 8, sometimes=True)
```

vojtapolasek:
@mildas  ok, I have accepted your suggestion.

PR #60, opened 2023-08-23
update waivers for ipv4_ip_forward audit_rules_privileged_commands

### LINKED ISSUE #58: allow matching on architecture as well
Currently, afaik we can match on RHEL version and scap-security-guide package version. Sometimes, a rule fails only on certain architecture.
Please add possibility to match the architecture.
Example productization issue: https://github.com/ComplianceAsCode/content/issues/10938

comps:
(From latest productization run.)

The `arch` addition fixes #58 even if it was ultimately not useful/necessary. Tested, works.


mildas (changes requested):
Please add also `sshd_use_strong_macs` `unknown` result waiver.

PR #66, opened 2023-08-29
remove `enable_fips_mode waiver` as it passes on latest RHEL9 compose

mildas:
In last weekly productization, `enable_fips_mode` rule fails because of `pass` result
```
fail: enable_fips_mode (waive: expected fail/error, got pass) 
```
I did couple of re-runs on latest 9.3 composes and still got same result.

Only recent SSG `enable_fips_mode` rule update I found is https://github.com/ComplianceAsCode/content/pull/11026 but it shouldn't affect remediation. Thus, I guess the change comes from compose update.

### DIFF
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -197,10 +197,6 @@
 /hardening/ansible/.+/mount_option_home_nosuid
     Match(True, sometimes=True)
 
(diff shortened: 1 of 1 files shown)


cortesana (approved):
LGTM

PR #67, opened 2023-08-30
Add waivers to resolve ComplianceAsCode/content#11031

Mab879:
We need to wave these rules due to order of remediations.

### DIFF
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -14,6 +14,12 @@
 /hardening/oscap(/with-gui)?/[^/]+/configure_bashrc_exec_tmux
 /hardening/oscap(/with-gui)?/[^/]+/no_tmux_in_shells
 /hardening/oscap(/with-gui)?/[^/]+/configure_usbguard_auditbackend
+/hardening/oscap(/with-gui)?/[^/]+/configure_tmux_lock_after_time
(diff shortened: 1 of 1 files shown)


conf/waivers-upstream
```diff
@@ -18,6 +18,9 @@
 /hardening/oscap(/with-gui)?/[^/]+/configure_bashrc_tmux
 /hardening/oscap(/with-gui)?/stig(_gui)?/postfix_prevent_unrestricted_relay
     rhel == 8
+/hardening/oscap(/with-gui)?/[^/]+/configure_tmux_lock_after_time
+/hardening/oscap(/with-gui)?/[^/]+/configure_tmux_lock_command
+/hardening/oscap(/with-gui)?/[^/]+/configure_tmux_lock_keybinding
```
mildas, line 23: Wrong syntax. After regex lines there must be a condition, see https://github.com/RHSecurityCompliance/contest/blob/main/WAIVERS.md

mildas (changes requested):

mildas (changes requested):
Now the waivers look good. But looking at productization results, not only `/hardening/oscap`  is affected, but also `/hardening/host-os` so please add the waivers there as well (few lines below your changes are very similar waivers for the other test)

PR #70, opened 2023-09-11
Replace waivers-released with waivers-upstream.

mildas:
0.1.69 will be released in downstream - upstream waivers will be applicable to released versions.
Few waivers are not included in waivers-released because they are not relevant for downstream as they are only in latest upstream.
Waivers that are not included in downstream waivers:
```
/hardening/oscap(/with-gui)?/[^/]+/configure_tmux_lock_after_time
/hardening/oscap(/with-gui)?/[^/]+/configure_tmux_lock_command
/hardening/oscap(/with-gui)?/[^/]+/configure_tmux_lock_keybinding
/hardening/host-os/oscap/[^/]+/configure_tmux_lock_after_time
/hardening/host-os/oscap/[^/]+/configure_tmux_lock_command
/hardening/host-os/oscap/[^/]+/configure_tmux_lock_keybinding
/hardening/.+/e8/sshd_use_strong_macs
```

### DIFF
--- a/conf/waivers-released
+++ b/conf/waivers-released
@@ -3,6 +3,11 @@
 # - see https://github.com/ComplianceAsCode/content/issues/9746
 /hardening/anaconda(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
     rhel >= 8
+# similar case for RHEL 9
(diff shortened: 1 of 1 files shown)


matusmarhefka (approved):
LGTM

PR #80, opened 2024-01-05
treat CentOS Stream as a valid versions.rhel match

comps:
In case you're wondering about the `is_true_rhel` naming, I didn't want
```python
if rhel.is_rhel:
```
as that seemed a bit counter-intuitive, compared to
```python
if rhel.is_true_rhel:
```
which hints at the RHEL/CentOS duality.


comps:
Tested on cs9, seems to work.

README.md
```diff
@@ -53,6 +53,19 @@ on Red Hat Enterprise Linux.
...
+
+- `CONTEST_PLAYBOOK`
+  - Specify a filesystem path to an Ansible YAML file to use as a playbook
+    for testing.  
+    A magical string of '{PROFILE}' expands to a profile name being tested.
+
+- `CONTEST_KICKSTART`
+  - Specify a filesystem path to an configuration file to use as an Anaconda
+    kickstart for testing.  
+    A magical string of '{PROFILE}' expands to a profile name being tested.
```
mildas, line 67: Magical string for distro major version , `DISTRO_MAJOR`, would be also useful.

lib/ansible.py
```diff
@@ -0,0 +1,31 @@
...
+    if versions.rhel.is_true_rhel():
+        # RHEL-7 seems to have it bundled in the base RPM
+        if versions.rhel <= 7:
+            return
+        pkg = 'rhc-worker-playbook'
+        ret = util.subprocess_run(['rpm', '--quiet', '-q', pkg])
+        if ret.returncode != 0:
+            util.subprocess_run(
+                ['dnf', '-y', '--nogpgcheck', 'install', pkg],
+                check=True)
```
mildas, line 20: Why not to leave it in `recommend` package?

lib/ansible.py
```diff
@@ -0,0 +1,31 @@
...
+        # https://access.redhat.com/articles/remediation
+        os.environ['ANSIBLE_COLLECTIONS_PATH'] = \
+            '/usr/share/rhc-worker-playbook/ansible/collections/ansible_collections/'
+
+    # CentOS and Fedora don't
+    else:
+        for collection in ['community.general', 'ansible.posix']:
+            util.subprocess_run(
+                ['ansible-galaxy', '-vvv', 'collection', 'install', collection],
+                check=True)
```
mildas, line 31: CentOS7, same as RHEL7, `ansible` package has all available Ansible modules (including `community.general` and `ansible.posix` modules).

README.md
comps: I thought about it as well, but it's not as straightforward.

I decided not to do it, because it would make the neat
```python
return Path(override.format(PROFILE=profile))
```
a lot longer, but mainly because you can already anticipate the distro major version before you run the tests, whereas, without `{PROFILE}`, you would have to run each test separately.

Finally, you already need to hardcode distro major version when running `tmt`, because of ie. `-c distro=rhel9` as context.

If you think it would still be useful despite these shortcomings, I can add it. Probably for datastream too then.

lib/ansible.py
comps: I guess that's a good point.

The `install_deps()` function can probably still remain there, so test code doesn't have to `if` based on distro type and version, but the RHEL section can just `rpm -q` check for `rhc-worker-playbook` to be installed and `RuntimeError()` if it is not.

Would that make sense?

lib/ansible.py
comps: Right, I'll make this applicable for 8+ then.

README.md
mildas: The TMT context point convinced me. Let's leave it as it is without distro major version.

lib/ansible.py
mildas: +1, makes sense.
I like the `rpm -q` check to get error right away. And not spending couple of minutes (e.g. on VM install) and then getting Ansible fatal error because of missing module.

mildas (approved):

PR #85, opened 2024-01-11
Various changes to make Ansible host-os tests ready for Github CI (CentOS)

mildas:
Testing Farm repos are not gpg signed, thus unselect those rules from remediations and waive their results.

Enable `/host-os/ansible` as we want to run it in Github CI (that's the only option how to test Ansible remediations in upstream, VM-based tests are not allowed). For that, update also waivers to include `host-os/ansible`.

Bugfixes:
- final scan is performed with final datastream, not with custom datastream where rules might be unselected - to see those rules in final scan (waivers for such rules are expected)
- use full profile name in test to avoid multiple profile matches

### DIFF
--- a/conf/waivers-released
+++ b/conf/waivers-released
@@ -1,7 +1,7 @@
 # remediation for this is globally disabled in Contest,
 # see conf/remediation.py
 /hardening/.*/accounts_password_set_max_life_existing
-    True
+    Match(True, sometimes=True)
 
 # requires running firewalld (firewall-cmd) and NetworkManager,
 # which are not available in their final form in the Anaconda environment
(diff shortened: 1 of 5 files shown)


conf/waivers-upstream
```diff
@@ -159,32 +159,32 @@
 # TODO: completely unknown, investigate and sort
 #
 # all RHELs
-/hardening/ansible/.+/mount_option_boot_noexec
+/hardenings/ansible/.+/mount_option_boot_noexec
```
comps, line 162: Probably a typo?

hardening/host-os/ansible/main.fmf
```diff
@@ -7,25 +7,20 @@ duration: 1h
...
-# VM-using tests, /hardening/ansible, and we would prefer the waivers for
-# those tests to be categorized or bugs fixed, before we extend the test
-# matrix with this host-os ansible test.
-enabled: false
+  - when: distro == rhel-7 or distro == centos-7
+    require+:
+      - ansible
+  - when: distro >= rhel-8 or distro >= centos-8
+    require+:
+      - ansible-core
```
comps, line 23: Would it make sense to copy/paste this change (adjust instead of recommends) to other Ansible tests?

mildas: I don't think it's worth it as quite soon we are going to get rid of RHEL7 completely and `ansible` requirement won't be relevant anymore

conf/waivers-released
```diff
@@ -1,7 +1,7 @@
 # remediation for this is globally disabled in Contest,
 # see conf/remediation.py
 /hardening/.*/accounts_password_set_max_life_existing
-    True
+    Match(True, sometimes=True)
```
comps, line 4: I'm a bit surprised by this accidentally passing, but OK, it makes sense to have `sometimes=True` here in general, because us not running remediation doesn't mean it'll automatically fail.

PR #91, opened 2024-01-22
Add an Image Builder test and a related library

comps:
(Probably go commit-by-commit.)
(Note: "osbuild" is the technology Image Builder is built on, FYI)

Most of the changes are actually reworking existing `virt` and `util` so we don't duplicate code with `osbuild`, changing VM guest customizations from a firstboot script to a RPM scriptlet.

I chose RPM because it's compatible with both Anaconda (oscap/anaconda/ansible remediations) and osbuild (image-builder), and it's literally the only way to do arbitrary OS customizations in an osbuild image.

* I plan to further refine the code, and change the Blueprint approach to generate a baseline from `oscap xccdf generate fix --fix-type blueprint`.
* I also want to do categorize random `virt` functions into `class Host` and `class Guest`, like I've done with `lib/osbuild.py`.
* I'm also redoing repository handling, adding them in the VM via RPM, because `[[customizations.repository]]` doesn't allow `file://` in `gpgkey=`, it needs HTTP.
* And probably more

But the current code is a GEFN base for these future changes, so pls review.


comps:
Note that I'm aware that `PARTITIONS` in `class Kickstart` (or the original text block version) are another example of "random hacks in tests" similar to installing RPM packages to make remediation work.

We basically pre-create partitions we know the content uses (whitebox testing).

For Image Builder, that particular commit will be useless, because the `oscap`-generated Blueprint contains partition sections, but I feel separating partitions from the rest of the text block is still an improvement, and makes it easier to move them to `conf` (or wherever) in the future.

STYLE.md
```diff
@@ -290,3 +290,7 @@ Misc:
   - pseudotty and other executables
   - ...
   - adjust `libdir` in `util` appropriately ?
+
+ThreadingHTTPServer in Python 3.8+ to replace the multiprocessing logic, and
```
mildas, line 294: Nit: Python 3.7+

lib/virt.py
```diff
@@ -403,19 +370,33 @@ def install(self, location=None, kickstart=None, ks_verbatim=False):
...
+        pack.post.append(GUEST_SETUP)
+        pack.requires += GUEST_SETUP_REQUIRES
+
+        srv = util.BackgroundHTTPServer(NETWORK_HOST, http_port)
+
+        with contextlib.ExitStack() as stack:
+            repo = stack.enter_context(pack.build_as_repo())
+            srv.add_file(repo, 'repo')
+            stack.enter_context(srv)
+            ksfile = stack.enter_context(kickstart.to_tmpfile())
```
mildas: Smart.
However, documentation is missing - in README, in function doc, or just a comment is fine.

lib/virt.py
```diff
@@ -382,23 +385,26 @@ def install(self, location=None, kickstart=None, ks_verbatim=False):
...
-        pack = util.RpmPack()
-        pack.post.append(self.SETUP)
-        pack.requires += self.SETUP_REQUIRES
-
-        srv = util.BackgroundHTTPServer(NETWORK_HOST, http_port)
-
         with contextlib.ExitStack() as stack:
+            pack = util.RpmPack()
+            pack.post.append(self.SETUP)
+            pack.requires += self.SETUP_REQUIRES
```
mildas, line 401: Why moving it to the context?

hardening/host-os/ansible/main.fmf
```diff
@@ -4,6 +4,8 @@ result: custom
 environment+:
     PYTHONPATH: ../../..
 duration: 1h
+require+:
+  - openscap-scanner
```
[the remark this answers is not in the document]
comps: I mean - having `host-os/main.fmf` crossed my mind later after creating this PR and it's definitely something we could do. It hides the require a little bit (people don't always look in parent dirs for `.fmf`), but it also deduplicates code.

OTOH it makes the tests a bit harder to copy/move around or use as templates for new tests.

Honestly, it's up to you, I can move the require if needed.

STYLE.md
```diff
@@ -196,6 +196,58 @@ with workarounds that would use operators.
...
+```
+my_list = [
+    1,
+    2,
+    ]
+ret = func_call(
+    some_long_arg="probably useless long text that should be in a variable",
+    another_arg=123,
+    )
+```
```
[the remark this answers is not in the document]
comps: Alright, I can remove it.

lib/util/dedent.py
```diff
@@ -0,0 +1,24 @@
...
+This allows raw blocks like
+
+    variable = dedent(fr'''
+        some
+        content
+    ''')
+
+without the leading newline, any common leading whitespaces, and the trailing
+newline with 4 spaces. You might think using '''\ would eliminate the first
+newline, but the string is 'raw', it doesn't have escapes.
```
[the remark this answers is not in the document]
comps: The `4` refers specifically to the example given,
```diff
+    variable = dedent(fr'''
+        some
+        content
+    ''')
```
which uses 4 spaces to indent the last line.

The "trailing newline" mention is important - it's the main reason for this function.

```python
def a():
    variable = fr'''
        some
        content
    '''
    return variable
```
results in
```
>>> a()
'\n        some\n        content\n    '
```
quoting the comment:
- without the leading newline (`\n`)
- any common leading whitespaces (`        ` before `some` and `content`)
- and the trailing newline (`\n`)
- with 4 spaces (`    ` after the last `\n`)

If it was too confusing, I can probably reword it, or maybe add the one-line result to better explain what problem the function solves. Or maybe encase the block inside `def a()` to avoid confusion with module-level (zero indent) `variable = ` definition.

lib/util/dedent.py
```diff
@@ -0,0 +1,24 @@
...
+
+import textwrap
+
+
+def dedent(text):
+    """
+    Like textwrap.dedent(), but also strip leading and trailing spaces/newlines
+    up to the content.
+    """
+    return textwrap.dedent(text.lstrip('\n').rstrip(' \n\t'))
```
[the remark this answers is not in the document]
comps: Huh, it actually does:
```
./lib/virt.py:239:10: E223 tab before operator
./lib/virt.py:239:10: W291 trailing whitespace
```
Okay, I can remove the `\t`.

lib/util/httpsrv.py
```diff
@@ -8,16 +27,36 @@
 
 
 class _BackgroundHTTPServerHandler(SimpleHTTPRequestHandler):
+    def send_file(self, path):
+        try:
+            with open(path, 'rb') as f:
+                self.send_response(200)
+                self.end_headers()
+                shutil.copyfileobj(f, self.wfile)
```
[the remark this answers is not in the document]
comps: No, HTTP first sends headers and then data. The headers typically contain `Content-Length`, so browsers (and other clients) can display a nice progress bar of % downloaded, or just tell how much data to read from the socket.

```
$ curl -v localhost:8000/passwd
*   Trying 127.0.0.1:8000...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> GET /passwd HTTP/1.1
> Host: localhost:8000
> User-Agent: curl/8.0.1
> Accept: */*
> 
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Server: SimpleHTTP/0.6 Python/3.11.4
< Date: Tue, 23 Jan 2024 17:13:54 GMT
< Content-type: application/octet-stream
< Content-Length: 32
< Last-Modified: Tue, 23 Jan 2024 17:13:31 GMT
< 
root:x:0:0:root:/root:/bin/bash
* Closing connection 0
```
(FTR, this block of code was already there, I just moved it to a dedicated function)

lib/util/httpsrv.py
```diff
@@ -1,3 +1,22 @@
...
+    srv.add_file('/on/disk/file.txt', '/visible.txt')
+    srv.add_file('/on/disk/dir', '/somedir')
+    with srv:
+        ...
+
+Any HTTP GET requests for '/visible.txt' will receive the contents of
+'/on/disk/file.txt' (or 404).
+
+Any HTTP GET requests for '/somedir/aa/bb' will receive the contents of
+'/on/disk/dir/aa/bb' (or 404).
```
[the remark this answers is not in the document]
comps: I'm not super happy about the implementation either, but adding proper directory index handling would have been a lot more code.

I can indeed at least expand the documentation like you suggested.

lib/util/httpsrv.py
```diff
@@ -8,16 +27,36 @@
...
+            # exact match = file
+            if get_path == map_path:
+                self.send_file(fs_path)
+                return
+            # common prefix = directory
+            common = os.path.commonpath([get_path, map_path])
+            if common not in ['', '/']:
+                relative = Path(get_path).relative_to(common)
+                self.send_file(fs_path / relative)
+                return
```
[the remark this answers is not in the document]
comps: No, that's a bug that my simplistic implementation didn't account for. I'll fix it.

STYLE.md
comps: Will adjust.

lib/virt.py
comps: I can add commented sections for each of `RpmPack`, `Kickstart`, and more if needed, but I probably wouldn't explain the behavior of standard Python modules - ie. we don't explain `NamedTuple` or ie. `ChainMap` either.

lib/virt.py
comps: Because it seemed clearer and kept related lines together.

Adding some comments maybe explains it better:
```python
        with contextlib.ExitStack() as stack:
            # this is all RpmPack
            pack = util.RpmPack()
            pack.post.append(self.SETUP)
            pack.requires += self.SETUP_REQUIRES
            repo = stack.enter_context(pack.build_as_repo())

            # this is all HTTP server
            srv = util.BackgroundHTTPServer(NETWORK_HOST, http_port)
            srv.add_file(repo, 'repo')
            stack.enter_context(srv)

            # this is all Kickstart
            ksfile = stack.enter_context(kickstart.to_tmpfile())
```
Yes, it means that if `BackgroundHTTPServer` fails to initialize an instance, `pack.build_as_repo()` has wasted 1 second doing extra work that will be undone, but that should never happen in practice - instance initialization is not listening on a socket, which would be done after repo building anyway in the old code:
```python
        with contextlib.ExitStack() as stack:
            repo = stack.enter_context(pack.build_as_repo())
            srv.add_file(repo, 'repo')
            stack.enter_context(srv)
            ksfile = stack.enter_context(kickstart.to_tmpfile())
```

hardening/host-os/ansible/main.fmf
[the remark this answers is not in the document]
mildas: Let's leave it as it is. At least for now. If there would be more tests where we could put requirements (or something else), I'd consider this. But now, as it would be the only case, it would rather complicate things than make them simpler.

lib/util/dedent.py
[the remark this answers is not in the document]
mildas: > The "trailing newline" mention is important - it's the main reason for this function.

Yes, I probably wrongly worded my comment. The "trailing newline" is important to mention, I agree.

> The 4 refers specifically to the example given,

That is my misunderstanding. When I read those sentences, I thought they are for the function in general, not for the example. To clarify my thoughts, when I read it (I was reading docstring first, so at the moment I didn't know what the function code looks like) I thought:
> Why it removes ONLY 4 spaces? What if there is another indentation level? It will leave trailing spaces

So I checked the code and saw it removes all trailing spaces.

lib/util/httpsrv.py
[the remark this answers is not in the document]
mildas: Ok, I see.

lib/virt.py
mildas: Yes, definitely don't explain those, everyone can google that (as I did with the `ExitStack` :D). All I want documented is the idea of package added to installation which does the setup to allow login.

lib/virt.py
mildas: Alright, I'm fine with that.

The idea, or rather how I write context (and it might be wrong that I made up it myself), is to put as less as possible to context and everything what doesn't need to be in it should stay outside. But looking at it now (lines together with comments), your approach is more sexy.

lib/osbuild.py
```diff
@@ -0,0 +1,370 @@
...
+            for pkg in virt.Kickstart.PACKAGES:
+                blueprint.add_package(pkg)
+            # copy partitions too
+            for mountpoint, size in virt.Kickstart.PARTITIONS:
+                blueprint.add_partition(mountpoint, size*1024*1024)
+            # generate the ssh key on the same place as virt.Guest
+            util.ssh_keygen(self.ssh_keyfile_path)
+            with open(f'{self.ssh_keyfile_path}.pub') as f:
+                pubkey = f.read().rstrip()
+            blueprint.add_user('root', password=virt.GUEST_LOGIN_PASS, ssh_pubkey=pubkey)
```
mildas, line 291: Does it work? I expect `root` is there by default and thus customization won't add/change it.

That's what came to mind when seeing [4.8.6. Setting SSH key for existing users](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/composing_a_customized_rhel_system_image/creating-system-images-with-composer-command-line-interface_composing-a-customized-rhel-system-image#setting-ssh-key-for-existing-users_creating-system-images-with-composer-command-line-interface)

mildas: And the password won't work as now you add plaintext but hash is expected.

lib/osbuild.py
```diff
@@ -0,0 +1,370 @@
...
+            if profile:
+                blueprint.add_openscap(self.DATASTREAM, profile)
+                #excludes = conf.remediation_excludes.host_os  # TEMP, TODO
+                excludes = []
+                blueprint.add_openscap_tailoring(unselected=excludes)
+
+        http_port = 8091
+        disk_path = Path(f'{virt.GUEST_IMG_DIR}/{self.name}.img')
+
+        with contextlib.ExitStack() as stack:
```
mildas, line 302: Similarly as for `lib/virt.py`, please document it somewhere (rpm package with the setup hack, repo, package in blueprint) . Also comments as you wrote in https://github.com/RHSecurityCompliance/contest/pull/91#discussion_r1463714137 (`# this is RPM part`, `# this is HTTP server`) would be nice.

hardening/image-builder/test.py
```diff
@@ -0,0 +1,35 @@
...
+g.create(profile=profile)
+
+profile = f'xccdf_org.ssgproject.content_profile_{profile}'
+
+with g.booted():
+    # old RHEL-7 oscap mixes errors into --progress rule names without a newline
+    verbose = '--verbose INFO' if versions.oscap >= 1.3 else ''
+    redir = '2>&1' if versions.oscap >= 1.3 else ''
+    # RHEL-7 HTML report doesn't contain OVAL findings by default
+    oval_results = '' if versions.oscap >= 1.3 else '--results results.xml --oval-results'
```
mildas, line 23: We can skip the RHEL7 workarounds as we know there will be no IB for RHEL7.

lib/virt.py
comps: FTR; I agree with you in general, don't put things that don't need to be in the context block, and that's how it should be for single-object contexts, it's just that in this case, I think trading that principle for the extra readability is worth it.

hardening/image-builder/main.fmf
```diff
@@ -0,0 +1,127 @@
...
+  - daily
+
+/anssi_bp28_high:
+    environment+:
+        PROFILE: anssi_bp28_high
+    adjust+:
+      - when: distro == rhel-7
+        enabled: false
+        because: RHEL-7 has anssi_nt28_high instead
+    extra-summary: /CoreOS/scap-security-guide/hardening/image-builder/anssi_bp28_high
```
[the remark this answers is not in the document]
comps: I was considering that, but my reasoning was that we have these blocks for all other hardening tests, and it might be easier to just leave them the same, so we can modify all of them in the same manner when removing RHEL-7.

But I have no issues making it RHEL8+ only for `image-builder` either, no strong opinions here.

lib/osbuild.py
comps: It does indeed work. I used to have a password crypt function (as evidenced by a leftover comment I'll remove during cleanup/documentation in later commits), but then I saw Blueprint examples with plaintexts and I switched back.

Then I read the docs on https://www.osbuild.org/guides/image-builder-on-premises/blueprint-reference.html#additional-users and it said
```
Note: If the password starts with $6$, $5$, or $2b$ it will be stored
as an encrypted password. Otherwise it will be treated as a plain text
password.
```

As for the ssh key, I also asked osbuild developers, and was told that it does the same thing as `[[customizations.sshkey]]` which is an older customization method, and that both are internally merged during blueprint parsing, and was recommended to use `[[customizations.user]]`.

lib/osbuild.py
comps: Sure, will do.

hardening/image-builder/test.py
comps: See my comment about `main.fmf`, I applied the same logic here.

hardening/image-builder/main.fmf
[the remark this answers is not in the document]
mildas: Fine, leave it there for now so it's same with other hardening tests (applies `/hardening/image-builder/tests.py` as well).

Just don't forget to limit distro in in `adjust`.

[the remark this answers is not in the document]
comps: I believe the is no need to change `adjust`, it should already be set up correctly from other tests (that I copied the `.fmf` from).

comps:
I've force-pushed fixes for the review, please see https://github.com/RHSecurityCompliance/contest/compare/fef287aa57345e312b70b0782661c0dd249cf888..9249133d125ea42f9dadb93f102cb4ca2f2855a8 with the diff. Aside from the last commit ([clean up lib.osbuild, add documentation](https://github.com/RHSecurityCompliance/contest/pull/91/commits/9249133d125ea42f9dadb93f102cb4ca2f2855a8)), everything has been amended to the original commits.

comps:
Well, Github added my second push to the list of commits from the first one. Thanks, Github.

Anyway, the second push contained extra commits above the fixes for the review.  
The extra commits are:
* [add repositories via RpmPack in lib.osbuild](https://github.com/RHSecurityCompliance/contest/pull/91/commits/29297437273da76b3a64fdae24a97e43f4b23f4f)
* [use relative imports in httpsrv and rpmpack](https://github.com/RHSecurityCompliance/contest/pull/91/commits/f0d1577bb077008f9f060db7cbc773d8ff052562)
* [use oscap-generated blueprints for lib.osbuild](https://github.com/RHSecurityCompliance/contest/pull/91/commits/748c9ec3a1e12f5f22de35abe95d6275c1adaae2)
* [clear out partitions for content-provided kickstarts](https://github.com/RHSecurityCompliance/contest/pull/91/commits/bce6c8800baf8b6b8d3807950346d99e98fb7bd7)
* [add with-gui variants of /hardening/image-builder](https://github.com/RHSecurityCompliance/contest/pull/91/commits/0d684eae80a8ba5a20bcca2e8e2de5f2857f92b3)
* [move lib.virt host setup into class Host](https://github.com/RHSecurityCompliance/contest/pull/91/commits/c311f2cc922063eabcc27247187c90188ed86240)

comps:
With these commits, I think I'm done with this PR (as in - will address review comments, but won't add more commits).

The IB tests are still disabled and without waivers, because, honestly, like 98% of them error out (only very few profiles work, only on RHEL-9, only non-gui) and we would be likely wasting machine time (and real time waiting for timeouts) running the tests in productization.

So the next step would be to figure out why DHCP isn't enabled on the `with-gui` package set (and `anssi_bp28_high`) and try upstream content, to see if things are any better. But otherwise, I see the newly-added test as an ad-hoc executed tool for now, unfortunately.

Turns out openscap in Image Builder is in a much, much worse shape than we thought, because official docs are wrong and didn't actually lead to properly remediated installs.

lib/osbuild.py
```diff
@@ -214,6 +214,12 @@ def add_package(self, name):
             name = "{name}"
         ''') + '\n'
 
+    def add_package_group(self, name):
+        self.assembled += util.dedent(fr'''
+            [[groups]]
+            name = "{name}"
+        ''') + '\n'
```
mildas, line 221: [4.7. RHEL image builder blueprint format](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/composing_a_customized_rhel_system_image/creating-system-images-with-composer-command-line-interface_composing-a-customized-rhel-system-image#composer-blueprint-format_creating-system-images-with-composer-command-line-interface)
```
[[groups]]
name = "group-name"
```
> The group entry describes a group of packages to be installed into the image. 

And [4.8.2. Selecting a package group](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9/html/composing_a_customized_rhel_system_image/creating-system-images-with-composer-command-line-interface_composing-a-customized-rhel-system-image#selecting-a-package-group_creating-system-images-with-composer-command-line-interface)
```
[[packages]]
name = "package_group_name"
```
>  Replace "package_group_name" with the name of the group. For example, "tmux". 

I don't see the difference. But just looking at function name and RHEL doc examples, the latter seems as the right one.

comps: I'll ask osbuild developers, maybe one of them is obsolete now.

comps: Turns out `[[groups]]` is the correct syntax (as I'm using it currently in the code), using groups in `[[packages]]` should not work and the documentation is probably wrong.

lib/virt.py
```diff
@@ -283,6 +264,11 @@ def append(self, content):
         """Append arbitrary string content to the kickstart template."""
         self.appends.append(content)
 
+    def add_pre(self, content):
+        new = ('%pre --interpreter=/bin/bash --erroronfail\n'
+               'set -xe; exec >/dev/tty 2>&1\n' + content + '\n%end')
+        self.append(new)
+
```
mildas, line 271: The function is not used anywhere

hardening/image-builder/main.fmf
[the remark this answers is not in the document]
mildas: I meant
```
# TODO: RHEL-8+ only because osbuild is not on RHEL-7
```
few lines above.

lib/virt.py
comps: I know, but as the commit message says:
```
This is mainly useful for debugging and patching Anaconda.
Copied from add_post().
```
I needed to add it for debugging `/hardening/anaconda` and, if we drop the commit, it's likely I'll need to re-create it next time we need to debug that test.

hardening/image-builder/main.fmf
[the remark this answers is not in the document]
comps: Ah, right, I guess there isn't anything to prevent it running on RHEL-7. I'll add an `adjust`.

comps:
Rebased on latest `origin/main`, resolved conflicts, disabled on RHEL-7:
```
  - when: distro == rhel-7
    enabled: false
    because: there is no OSBuild or Image Builder on RHEL-7
```

mildas (approved):

PR #100, opened 2024-02-09
Add imagebuilder waivers

comps:
(Probably go commit by commit.)

These waivers result in both RHEL-8 and RHEL-9, and both downstream (currently old `scap-security-guide` RPM) and upstream (latest CaC/content) somewhat working, in that there should be no `error` or `fail`.

I'd like to run these regularly (daily? weekly?) to catch improvements as they happen, both on the RHEL side with OSBuild, and on the content side as Github issues get fixed. As such, the tags were copied from `/hardening/anaconda`.

### DIFF
--- a/conf/waivers-released
+++ b/conf/waivers-released
@@ -1,13 +1,15 @@
 # requires running firewalld (firewall-cmd) and NetworkManager,
 # which are not available in their final form in the Anaconda environment
 # - see https://github.com/ComplianceAsCode/content/issues/9746
+#       https://github.com/ComplianceAsCode/content/issues/11318
 /hardening/anaconda(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
-    rhel >= 8
-# similar case for RHEL 9
+/hardening/image-builder(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
 # https://github.com/ComplianceAsCode/content/issues/10939
-/hardening/anaconda(/with-gui)?/[^/]+/firewalld_loopback_traffic_restricted
-/hardening/anaconda(/with-gui)?/[^/]+/firewalld_loopback_traffic_trusted
-    rhel >= 9
+# https://github.com/ComplianceAsCode/content/issues/10613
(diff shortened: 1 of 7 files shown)


comps:
Removed the hack for speeding up composer service restart, it was triggering random
```
ERROR: ComposePushErrored: No worker for arch 'x86_64'  available
```
much more frequently than they usually happen.

conf/waivers-released
```diff
@@ -2,12 +2,12 @@
...
-# similar case for RHEL 9
 # https://github.com/ComplianceAsCode/content/issues/10939
-/hardening/anaconda(/with-gui)?/[^/]+/firewalld_loopback_traffic_restricted
-/hardening/anaconda(/with-gui)?/[^/]+/firewalld_loopback_traffic_trusted
-    rhel >= 9
+# https://github.com/ComplianceAsCode/content/issues/10613
+/hardening/anaconda(/with-gui)?/[^/]+/firewalld_loopback_traffic_(restricted|trusted)
+# https://github.com/ComplianceAsCode/content/issues/11564
+/hardening/image-builder(/with-gui)?/[^/]+/firewalld_loopback_traffic_(restricted|trusted)
+    rhel >= 8
```
mildas, line 10: Do we want to mix IB here? Or should the IB waiver be among other IB waivers?

It's about adding rule to "rule specific section" vs "test specific section". This one you added close to existing Anaconda waiver (as it's the same rule), but for `firewalld_sshd_port_enabled` you added it to IB waivers section instead of to Anaconda waiver. Let's make it consistent.

comps: Hmm, good point.

I used a new section only for IB-specific fails, anything shared with other tests should probably be combined with the existing waivers. I'll move `firewalld_sshd_port_enabled` to the anaconda part.

Btw, I'll remove `firewalld_sshd_port_enabled` waivers completely (incl. Anaconda) from `waivers-upstream` and re-test everything, because I'm pretty sure the remediation for that rule has been removed, since we don't need an IB waiver either.

PR #104, opened 2024-02-20
Add `/per-rule` test for running `automatus.py` in rule mode

comps:
The rsync functionality was originally added for a custom Automatus runner implementation, and is currently unused, but I don't want to throw it away since it will be useful once CaC/content implements building templated/jinja2'd tests.

### DIFF
--- a/per-rule/test.py
+++ b/per-rule/test.py
@@ -1,0 +1,240 @@
+#!/usr/bin/python3
+
+import os
+import sys
+import re
+import subprocess
+import contextlib
+import tempfile
+import textwrap
+import distutils.dir_util
+from pathlib import Path
+
+from lib import util, results, oscap, virt, versions, dnf
+from conf import partitions
+
+
+@contextlib.contextmanager
+def get_content():
+    from_env = os.environ.get('CONTENT_SOURCE')
+    if from_env:
+        content = Path(from_env)
+        build_dir = content / 'build'
+
+        # if it has built content
+        if (build_dir / 'Makefile').exists():
+            util.log(f"using pre-built content: {build_dir}")
+            yield content
+
+        # manually build the source
+        else:
+            util.log(f"building content from source: {content}")
+            with tempfile.TemporaryDirectory() as tmpdir:
+                # TODO: this should be shutil.copytree(.., dirs_exist_ok=True) on python 3.8+
+                distutils.dir_util.copy_tree(content, tmpdir)
+                util.log(f"using {tmpdir} as content source")
+                # install dependencies
+                cmd = ['dnf', '-y', 'builddep', '--spec', 'scap-security-guide.spec']
+                util.subprocess_run(cmd, check=True, cwd=tmpdir)
+                # build content
+                cmd = ['./build_product', f'rhel{versions.rhel.major}']
+                util.subprocess_run(cmd, check=True, cwd=tmpdir)
+                yield Path(tmpdir)
+
+    else:
+        # fall back to SRPM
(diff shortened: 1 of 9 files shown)


per-rule/test.py
```diff
@@ -0,0 +1,265 @@
...
+                    ['rpm', '-q', '--qf', '%{NAME}-%{VERSION}', '-p', src_rpm],
+                    check=True, stdout=subprocess.PIPE, universal_newlines=True, cwd=tmpdir,
+                )
+                name_version = ret.stdout.strip()
+                extracted = Path(tmpdir) / 'BUILD' / name_version
+                util.log(f"using {extracted} as content source")
+                if not extracted.exists():
+                    raise FileNotFoundError(f"{extracted} not in extracted/patched SRPM")
+                # build content
+                # TODO: temporary, see https://github.com/ComplianceAsCode/content/pull/11606
```
mildas, line 67: PR is merged

mildas (commented):
Multiple `per-rule/test.py` functions seem to be a good fit for a `util/` - functions like `get_all_profiles()`, `get_all_rules()`, or `get_content()`. For now, they will be used only in this test, but I don't like them in `test.py` code. They seem as functions useful in general and that they could be used for other future beakerlib test re-implementations.

per-rule/test.py
comps: Yes, but we'll be running with unfixed content for a while, so the TODO needs to stay there until the next SSG release.

comps:
> Multiple `per-rule/test.py` functions seem to be a good fit for a `util/` - functions like `get_all_profiles()`, `get_all_rules()`, or `get_content()`. For now, they will be used only in this test, but I don't like them in `test.py` code. They seem as functions useful in general and that they could be used for other future beakerlib test re-implementations.

Yeah, I was actually thinking the same thing - if we want a replacement for `datastream-diff` (or whatever was it called), a function for listing all the profiles + function for listing all rules in a profile would come in handy.

I can move them to `lib.oscap`.

comps:
Moved the rule/profile functions.

I will probably do something with `get_content()` in a subsequent PR - I'm considering getting rid of `CONTEST_KICKSTARTS`, `CONTEST_DATASTREAM`, etc. in favor of one `CONTEST_CONTENT_SOURCE`, since that seems to be the use case. It's easier to just `scp` the whole CaC/content than digging out kickstarts and copying those to the target system.

comps:
One more thing I'm missing is an `adjust` for RHEL-8+ only. Will add that too.

per-rule/test.py
```diff
@@ -0,0 +1,239 @@
...
+                    status = 'info'
+                    note = 'not applicable'
+                elif status == 'found issue:':
+                    status = 'fail'
+                    note = f'profile:{profile}' if profile != '(all)' else None
+                else:
+                    status = 'error'
+                    note = "unknown Script status"
+
+                report_test_with_log(status, note, log_dir, rule_name, test_name)
```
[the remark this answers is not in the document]
comps: That's done on a higher level - without `CONTEST_VERBOSE`, only `fail` and `error` results are reported with their logs. Any other results are discarded, incl. their logs.

comps:
Made some small changes:
* `import distutils.dir_util` is needed on RHEL-8 (python3.8)
* `if (build_dir / 'Makefile').exists():` because CaC/content does include `build/.gitkeep`, which confused `.iterdir()`, so it always thought the user-provided content is already built
* `def slice_list(..)` fix for using it with python's list slicing

mildas (approved):

PR #105, opened 2024-02-21
add Nitrate test case IDs to /per-rule

comps:


mildas (changes requested):
I don't see it in general test plan, please export it also there.

mildas (approved):

PR #108, opened 2024-02-26
add waiver for firewalld_sshd_port_enabled during IB build on RHEL 8

vojtapolasek:
https://github.com/ComplianceAsCode/content/issues/11625


comps:
Please add the line to the block of lines above it, see: https://github.com/RHSecurityCompliance/contest/blob/1706c5cecba/conf/waivers-released#L6

comps:
Combined with the other firewalld sshd waivers, merging.

PR #109, opened 2024-02-27
Perform double remediation for `oscap` hardening

comps:
While making the feature, I needed to have working unselects, hence this PR being not very straightforward and having multiple somewhat-but-not-really related commits, sorry.

The goal of double `oscap` remediation is to get rid of many waivers which are a result of rule dependencies (`notapplicable` during first remediation, other applicable rules causing the condition to flip and the rule becoming applicable and failing during a scan -- second remediation run should catch and fix it).

Not a great solution, but likely the least bad.

[make /hardening/host-os/oscap reboot between two remediations](https://github.com/RHSecurityCompliance/contest/pull/109/commits/69dbb826721dc7e004c7c786af29cdfe14f3dd01) is not part of [run oscap-based remediation twice](https://github.com/RHSecurityCompliance/contest/pull/109/commits/8bdbe11baffb94a5f273c3165426ae636c3a0a06) because of the commit between them conflicting, and I didn't want to spend another hour re-doing all the commits just so I could squash the two. Hopefully it's not a big deal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ See also TODOs in [STYLE.md](STYLE.md), this is a temporary limitation.
 
 ### Virtual machines and logging in
 
-The VM-using `/hardening` tests do two hacks to allow login after hardening:
+The tests perform some hacks to allow login after hardening:
 
 - `-oPermitRootLogin=yes` in `OPTIONS` of `/etc/sysconfig/sshd`
   - This is to bypass ssh-denied root login. Doing this seems easier than trying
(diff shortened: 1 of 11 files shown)


mildas (commented):
Code lgtm. But I'm starting to question what and what not should be as variable - `scan-ds.xml`, `remediation-ds.xml`, `['remediation.html', 'remediation2.html']`, `report.html`, and probably more.
The naming is same across tests you changed but it might be different in other tests as they were not touched. Not a big deal (and maybe even not true, I haven't checked other tests). Single place, `const.py`, where everything would on single place could make it a slightly more readable though. Just food for thought.

mildas (approved):

comps:
> Single place, const.py, where everything would on single place could make it a slightly more readable though. Just food for thought.

I don't know about that. It would make the HTML reports more consistent, but the code likely less readable.

Compare this:
```python
    oscap.unselect_rules(util.get_datastream(), 'remediation-ds.xml', remediation.excludes())
    g.copy_to('remediation-ds.xml')

    # - remediate twice due to some rules being 'notapplicable'
    #   on the first pass
    for html_report in ['remediation.html', 'remediation2.html']:
```
to this:
```python
    # copy our datastream to the guest
    oscap.unselect_rules(util.get_datastream(), const.remediation_ds_file, remediation.excludes())
    g.copy_to(const.remediation_ds_file)

    # - remediate twice due to some rules being 'notapplicable'
    #   on the first pass
    for html_report in [const.remediation_html_1_file, const.remediation_html_2_file]:
```
To me, the string constants are easier to understand, because I don't have to look up what `const` is, what those variables are, etc.

At least in this specific case - in other cases when referring to the variable name is easier than referring to its value, I prefer the variable name:
```python
pi = 3.141592

result = pi * ...
```

PR #111, opened 2024-03-04
Remove chronyd_or_ntpd_set_maxpoll waiver

mildas:
The rule gets fixed by double remediation.

For further information see last comments in https://github.com/ComplianceAsCode/content/issues/11609


comps:
LGTM, I've seen the rule fail consistently in my runs (on passing), but haven't touched the waiver since it differed from nightly productization runs, where it apparently was not an issue (?)

PR #112, opened 2024-03-04
Add DISA alignment test for Bash remediation

mildas:
1. Harden the system using scap-security-guide Bash remediations (`oscap xccdf eval --remediate ...`)
2. Scan the system with our and DISA benchmark and get ARF results for comparing results
3. Compare results via `compare_results.py` scripts from scap-security-guide utils

The diff results reporting part (last `for`) will be same for all DISA Alignment tests (3x - Bash, Ansible, OAA). Is it worth of having it in `lib/`?

### DIFF
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -281,4 +281,22 @@
 /per-rule/[^/]+/tftpd_uses_secure_mode/wrong.fail
     rhel == 9
 
+# DISA Alignment waivers
+#
+# https://github.com/ComplianceAsCode/content/issues/9307 (DISA issue)
+/scanning/disa-alignment/oscap/sysctl_kernel_yama_ptrace_scope
+# https://github.com/ComplianceAsCode/content/issues/9308 (DISA issue)
+/scanning/disa-alignment/oscap/sysctl_kernel_core_pattern
+# https://github.com/ComplianceAsCode/content/issues/10044 (DISA issue)
+/scanning/disa-alignment/oscap/accounts_password_pam_pwhistory_remember_password_auth
+# https://github.com/ComplianceAsCode/content/issues/11197 (DISA issue)
+/scanning/disa-alignment/oscap/display_login_attempts
+    rhel == 8
+# https://github.com/ComplianceAsCode/content/issues/11649 (DISA issue)
+/scanning/disa-alignment/oscap/installed_OS_is_vendor_supported
+# https://github.com/ComplianceAsCode/content/issues/11650
+/scanning/disa-alignment/oscap/kernel_module_tipc_disabled
+    rhel == 9
+
+
 # vim: syntax=python
--- a/scanning/disa-alignment/oscap/test.py
+++ b/scanning/disa-alignment/oscap/test.py
@@ -1,0 +1,107 @@
+#!/usr/bin/python3
+import os
+import re
+import subprocess
+
+from lib import util, results, virt, oscap, versions
+from conf import partitions, remediation
+
+
+virt.Host.setup()
+
+profile = 'xccdf_org.ssgproject.content_profile_stig'
+
+g = virt.Guest('minimal_with_oscap')
+
+if not g.can_be_snapshotted():
+    ks = virt.Kickstart(partitions=partitions.partitions)
+    g.install(kickstart=ks)
+    g.prepare_for_snapshot()
+
+with g.snapshotted():
+    oscap.unselect_rules(util.get_datastream(), 'remediation-ds.xml', remediation.excludes())
+    g.copy_to('remediation-ds.xml')
+
+    for _ in range(2):
+        cmd = [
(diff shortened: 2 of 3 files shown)


hardening/disa-alignment/oscap/test.py
```diff
@@ -0,0 +1,103 @@
...
+if not g.can_be_snapshotted():
+    ks = virt.Kickstart(partitions=partitions.partitions)
+    g.install(kickstart=ks)
+    g.prepare_for_snapshot()
+
+with g.snapshotted():
+    oscap.unselect_rules(util.get_datastream(), 'remediation-ds.xml', remediation.excludes())
+    g.copy_to('remediation-ds.xml')
+
+    for html_report in ['remediation.html', 'remediation2.html']:
```
comps, line 24: You don't seem to be using these HTML reports - I would either add
```python
g.copy_from(html_report)
```
and add them to `logs=[]`, or (more realistically) just do
```python
for _ in range(2):
    ...
    # remove the --report html_report from oscap CLI
    ...
```

hardening/disa-alignment/oscap/test.py
```diff
@@ -0,0 +1,103 @@
...
+            compare_script, 'ssg-stig-viewer.xml', 'disa-arf.xml',
+        ]
+        _, lines = util.subprocess_stream(cmd, env=env)
+
+        # Different results format: CCE CCI - DISA_RULE_ID SSG_RULE_ID  -  SSG_RESULT - DISA_RESULT
+        result_regex = re.compile(r'[\w-]+ [\w-]+ - [\w-]+ (\w*)\s+(\w+) - (\w+)')
+        for line in lines:
+            result_match = result_regex.search(line)
+            if result_match:
+                # Use 1,2,3 = result_match.groups()
```
comps, line 91: You should follow this advice. :)

hardening/disa-alignment/oscap/test.py
```diff
@@ -0,0 +1,104 @@
...
+        env['PYTHONPATH'] = str(content_dir)
+        cmd = [
+            compare_script, 'ssg-stig-viewer.xml', 'disa-arf.xml',
+        ]
+        _, lines = util.subprocess_stream(cmd, env=env)
+
+        # Same result format:      CCE CCI - DISA_RULE_ID SSG_RULE_ID     RESULT
+        # Different result format: CCE CCI - DISA_RULE_ID SSG_RULE_ID     SSG_RESULT - DISA_RESULT
+        result_regex = re.compile(r'[\w-]+ [\w-]+ - [\w-]+ (\w*)\s+(\w+)(?: - *(\w+))*')
+        for match in result_regex.finditer('\n'.join(lines)):
```
comps, line 88: Okay, that was a bad advice with `re.finditer()`, I didn't realize the only input it has is a full string.

So `util.subprocess_stream()` does all the hard work of efficiently streaming the command output line-by-line, and that gets all thrown away by `.join()` just reading everything, merging it into one giant string, passing it to `.finditer()` to be broken down again and iterated over.

If the output is not too huge (ie. not Anaconda output), this might be okay, but the previous solution is IMHO better. Your call on which one you want to use.

comps: If you really want `re.finditer()`, then I would just do
```python
proc = util.subprocess_run(cmd, env=env)
```
above, and then just pass `proc.stdout.rstrip('\n')` to `re.finditer()` instead of `.join()`ing the lines together.

scanning/disa-alignment/oscap/test.py
```diff
@@ -0,0 +1,107 @@
...
+                rule_id = 'rule_id_not_found'
+            # Only 1 result matched - same results
+            if not disa_result:
+                results.report('pass', rule_id)
+            # SSG CPE checks can make rule notapplicable by different reason (package not
+            # installed, architecture, RHEL version). DISA bechmark doesn't have this
+            # capability, it just 'pass'. Ignore such result misalignments
+            elif ssg_result == 'notapplicable' and disa_result == 'pass':
+                result_note = 'SSG result: notapplicable, DISA result: pass'
+                results.report('pass', rule_id)
```
comps, line 101: You don't seem to be using the `result_note` in `results.report()`.

conf/waivers-upstream
```diff
@@ -281,4 +281,22 @@
...
+# https://github.com/ComplianceAsCode/content/issues/11197 (DISA issue)
+/scanning/disa-alignment/oscap/display_login_attempts
+    rhel == 8
+# https://github.com/ComplianceAsCode/content/issues/11649 (DISA issue)
+/scanning/disa-alignment/oscap/installed_OS_is_vendor_supported
+# https://github.com/ComplianceAsCode/content/issues/11650
+/scanning/disa-alignment/oscap/kernel_module_tipc_disabled
+    rhel == 9
+
+
```
comps, line 301: Probably remove the extra empty line here, we don't AFAIK use double newlines in waivers.

PR #117, opened 2024-03-12
Add new test /static-checks/html-links

matusmarhefka:
Test verifies that HTML links from datastream are accessible.

Rewrite of old `Sanity/oscap-html-links` test.

The new test also identified a broken link in datastreams, reported in https://github.com/ComplianceAsCode/content/issues/11678

### DIFF
--- a/static-checks/html-links/test.py
+++ b/static-checks/html-links/test.py
@@ -1,0 +1,25 @@
+#!/usr/bin/python3
+
+import re
+import requests
+
+from lib import util, results
+
+
+headers = {'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64;)'}
+url_regex = re.compile(r'href=[\'"]?(http[^\'" ]+)', re.IGNORECASE)
+
+with open(util.get_datastream(), 'r') as ds:
+    ds_content = ds.read()
+    urls = set(url_regex.findall(ds_content))
+
+for url in urls:
+    try:
+        r = requests.get(url, timeout=10, headers=headers)
+        r.raise_for_status()
+    except requests.exceptions.RequestException as err:
+        results.report('fail', url, err)
+    else:
+        results.report('pass', url)
+
+results.report_and_exit()
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -294,5 +294,14 @@
 /scanning/disa-alignment/oscap/kernel_module_tipc_disabled
     rhel == 9
 
+# HTML links from datastreams waivers
+#
+# Inaccessible until form is filled:
+/static-checks/html-links/https://docs-prv.pcisecuritystandards.org/PCI%20DSS/Standard/PCI-DSS-v4_0.pdf
+    True
+# https://github.com/ComplianceAsCode/content/issues/11678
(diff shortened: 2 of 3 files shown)


static-checks/html-links/test.py
```diff
@@ -0,0 +1,36 @@
...
+headers['User-Agent'] = 'Mozilla/5.0 (X11; Linux x86_64;)'
+url_regex = re.compile(r'href=[\'"]?(http[^\'" ]+)', re.IGNORECASE)
+
+with open(util.get_datastream(), 'r') as ds:
+    ds_content = ds.read()
+
+for url in set(url_regex.findall(ds_content)):
+    if url in problematic_urls:
+        results.report('skip', url, 'inaccessible by automated web scraping')
+        continue
```
mildas, line 26:
```suggestion
with open(util.get_datastream(), 'r') as ds:
    ds_content = ds.read()
    all_urls = url_regex.findall(ds.read())
    urls = set(filter(lambda x: x not in problematic_urls, all_links))

for url in urls:
```
what do you think about this? Checking in `for` only valid links

static-checks/html-links/test.py
```diff
@@ -0,0 +1,36 @@
...
+
+    req = urllib.request.Request(url, headers=headers)
+    try:
+        urllib.request.urlopen(req, timeout=10)
+    except (urllib.error.HTTPError, urllib.error.URLError, TimeoutError) as err:
+        results.report('fail', url, f'{err}')
```
mildas, line 32: `f'{err}'` is `err`, right?

static-checks/html-links/test.py
```diff
@@ -0,0 +1,36 @@
...
+
+# Following links are inaccessible because some websites detect automated web scraping
+# and return 'HTTP Error 403: Forbidden' or just hang indefinitely.
+problematic_urls = [
+    'https://www.cyber.gov.au/acsc/view-all-content/ism',
+    'https://docs-prv.pcisecuritystandards.org/PCI%20DSS/Standard/PCI-DSS-v4_0.pdf'
+]
+
+headers = {}
+headers['User-Agent'] = 'Mozilla/5.0 (X11; Linux x86_64;)'
```
mildas, line 17: I'd go with one-liner
```
headers = {'User-Agent': 'Mozilla/5.0 (X11; Linux x86_64;)'}
```

static-checks/html-links/test.py
mildas: Or maybe even better would be to not `filter` the `problematic_urls` and treat them like other links. Instead of maintaining such list in `test.py`, we would have them in waivers with other known issues.

static-checks/html-links/test.py
```diff
@@ -0,0 +1,36 @@
...
+        results.report('skip', url, 'inaccessible by automated web scraping')
+        continue
+
+    req = urllib.request.Request(url, headers=headers)
+    try:
+        urllib.request.urlopen(req, timeout=10)
+    except (urllib.error.HTTPError, urllib.error.URLError, TimeoutError) as err:
+        results.report('fail', url, f'{err}')
+
+    results.report('pass', url)
```
mildas, line 34: You report failures twice - once as `fail` and then as `pass`

mildas (approved):

PR #121, opened 2024-03-15
Upload `results-arf.xml`, other minor fixes

comps:
(See changes commit-by-commit.)

* `add /lib and /conf to test requires` is an old commit that got reverted last year because TestingFarm didn't have a new enough TMT. It now does, so it should work (and future-proof the repo from a compatibility breaking change on TMT's side, as they plan to stop copying all by default).
* `drop oscap verbose .txt logs` - the verbose logs were really not useful, so I replaced them with ARF XMLs


scanning/disa-alignment/shared.py
```diff
@@ -6,8 +6,6 @@
 profile = 'stig'
 profile_full = f'xccdf_org.ssgproject.content_profile_{profile}'
 
-# old RHEL-7 oscap mixes errors into --progress rule names without a newline
-redir = '2>&1' if versions.oscap >= 1.3 else ''
```
mildas, line 10: Isn't it still valid?

comps, line 10: FTR; This was there to combine (mix) `--progress` with the verbose output, so we could split the verbose output per-rule, using the `--progress` output as delimiters.

Since we don't do that anymore (no verbose output), we can just eat stdout, which contains pure `--progress`, and leave stderr on the console, if any warnings/errors are printed.

PR #126, opened 2024-03-21
waive service_sssd_enabled

vojtapolasek:
this rule requires additional manual sssd configuration which can't be attained by our rules in a meaningful way


comps:
I think you're missing a second line to the waiver,
```
lib.waive.WaiveParseError: waiver line 4: unexpected empty line between regexes
```

Issue #124, opened 2024-03-20
Support testing CaC/content and openscap build from pull requests


comps:
* For https://github.com/ComplianceAsCode/content , it means extending [`/plans/upstream`](https://github.com/RHSecurityCompliance/contest/blob/06a8a6ba0/plans/upstream.fmf) to support `--branch`, or rather to rewrite `git-clone` to `git-init` + `git-fetch`, so we can specify a non-branch `ref` such as `refs/pull/1234/head`, effectively cloning (and building in `get_content()`) a content snapshot from PR #1234.
* For openscap, it would mean extending the same plan (or `/plans/main.fmf` for all plans) to download `https://copr.fedorainfracloud.org/coprs/packit/OpenSCAP-openscap-{PR_ID}/repo/centos-stream-{versions.rhel.major}/packit-OpenSCAP-openscap-{PR_ID}-centos-stream-{versions.rhel.major}.repo` to `/etc/yum.repos.d` and calling `dnf upgrade openscap`.
  * Since this adds an HTTP-style git repo (not a `file://` one), the repo will be automagically used by all tests that propagate HTTP repos to VMs, incl. `/hardening/{oscap,ansible,image-builder}`.

PR #127, opened 2024-03-21
add CONTEST_CONTENT_PR and CONTEST_OSCAP_PR

comps:
Fixes https://github.com/RHSecurityCompliance/contest/issues/124

### DIFF
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -2,15 +2,66 @@
 /:
     inherit: false
 
+discover:
+    how: fmf
+
+execute:
+    how: tmt
+
 adjust:
-  - prepare:
+  - prepare+:
       - how: shell
         name: install-epel-on-rhel7
         because: python3 is not on RHEL 7 by default
         script: |
+            set -xe
             if ! rpm -q epel-release; then
-                curl -o epel-release.rpm --retry 10 https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+                curl --retry 10 -sSfkL -o epel-release.rpm https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
                 rpm -ivh epel-release.rpm
                 rm -f epel-release.rpm
             fi
     when: distro < rhel-8
+
+  - prepare+:
+      - how: shell
+        name: Update OpenSCAP from Packit (if specified)
+        script: |
+            set -xe
+            [ -n "$CONTEST_OSCAP_PR" ] || exit 0
+            repofile=/etc/yum.repos.d/openscap-packit-pr.repo
+            rm -f "$repofile"
+            if rpm -q openscap-scanner; then
+                # the RPM might have been upgraded from another PR, sanitize it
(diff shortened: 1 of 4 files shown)


plans/upstream.fmf
```diff
@@ -1,22 +1,22 @@
...
+        content_dir=/root/upstream-content
+        [ -e "$content_dir" ] || git init "$content_dir"
+        cd "$content_dir"
+        git fetch --depth=1 https://github.com/ComplianceAsCode/content.git "$ref"
+        # remove any previously built content
+        git clean -ffdx
+        git checkout FETCH_HEAD
 
-adjust+:
-  - prepare+:
```
comps, line 9: FYI: This is just moving all the code from `adjust` to root-level `prepare`, since there was no need for the `adjust` (no condition specified).

plans/main.fmf
```diff
@@ -2,15 +2,66 @@
...
 
+discover:
+    how: fmf
+
+execute:
+    how: tmt
+
 adjust:
-  - prepare:
+  - prepare+:
```
mildas, line 12: Why `+`? It doesn't inherit from top level, leaving `prepare:` would make it clean for reader no additional prepare steps coming from top.

plans/main.fmf
```diff
@@ -2,15 +2,66 @@
...
       - how: shell
         name: install-epel-on-rhel7
         because: python3 is not on RHEL 7 by default
         script: |
+            set -xe
             if ! rpm -q epel-release; then
-                curl -o epel-release.rpm --retry 10 https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
+                curl --retry 10 -sSfkL -o epel-release.rpm https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
mildas, line 19: I'd prefer long form options

plans/main.fmf
comps: I remember thinking the same and testing it, and it removed the other `prepare` content.

What I think is happening is that `adjust` is being applied **on top** of regular dictionary keys, so if `/plans/foobar.fmf` specifies `prepare:`, it gets used first, and then all the `adjust`s are applied, starting from the top-level `/plans/main.fmf`.

plans/main.fmf
comps: Generally, I do too, but we use curl short options on many places (at least internally) to the point where it became sort of a habit. As in - one should never use `curl` without these (for "gimme what's on that URL" style of use cases).

Writing it out would be a lot longer and need line breaks,
```
curl --output epel-release.rpm --retry 10 --silent --show-error --fail --insecure --location https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
especially `-o` (as an alternative to `wget`) I've never seen expanded to `--output`.

I modified this line only because I added the extra `curl` options elsewhere and thought it's a good idea to keep the options consistent across plans.

So I personally prefer the short options for `curl`, but I can use the long ones if it matters to you.

PR #129, opened 2024-03-26
The rule accounts_polyinstantiated_tmp appears to be fixed

Mab879:
Since it is fixed we need to remove the wavier.

### DIFF
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -171,7 +171,6 @@
 # RHEL-9 only
 /hardening(/host-os)?/ansible/.+/dnf-automatic_apply_updates
 /hardening(/host-os)?/ansible/.+/dnf-automatic_security_updates_only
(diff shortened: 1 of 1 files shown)


comps:
This is probably due to https://github.com/ComplianceAsCode/content/pull/11626/commits/13d2d865216a1b41552bfeb239414a8e8bde863e:
```diff
-# platform = multi_platform_fedora,Red Hat Enterprise Linux 7,Red Hat Enterprise Linux 8,multi_platform_sle
+# platform = multi_platform_fedora,multi_platform_rhel,multi_platform_sle
```
@jan-cerny would it make sense to fix this for the other ansible rules as well (make them RHEL-9 relevant?) .. I've looked into `accounts_polyinstantiated_var_tmp` and it seems to have the same issue, and chances are that others from the waivers,
```
dnf-automatic_apply_updates
dnf-automatic_security_updates_only
accounts_polyinstantiated_var_tmp
force_opensc_card_drivers
```
do too.

Issue #87, opened 2024-01-16
Differentiate remediation types in `has_no_remediation()`


comps:
Currently, the function looks for Bash remediations only - this may incorrectly waive failing Ansible results that do have a valid Ansible remediation, but no Bash remediation scripts.

Also, the existing Bash-based logic is fairly hacky and manual. Try parsing the datastream XML instead.

mildas:
For searching Ansible remediation, we could utilize similarly as for Bash `# oscap xccdf generate --profile '(all)' fix --fix-type ansible` and in output search for rule id. Ansible remediation has its rule id in `tags:`.

However, this approach will not work for IB as `--fix-type blueprint` creates you blueprint without any indication what rule is covered by what.

comps:
Significant help towards this is https://github.com/RHSecurityCompliance/contest/pull/155/commits/e1100f0428ef43b8eb41ffa9ee3719c2235e1e60 done via https://github.com/RHSecurityCompliance/contest/pull/155 .

This allows us to directly query bash/ansible/etc. remediations in the datastream XML.

Now we'd need to figure out how `has_no_remediation()` should work and how / if to parametrize it based on the remediation type being tested (`/hardening/oscap` vs `/hardening/ansible` vs ...).

comps:
The main reason we need to do this is to fix some false positives on non-Bash remediation technologies.

For example - kickstart / OAA / image-builder hardening **should** remediate some things that oscap alone cannot, like disk partitions (ie. by including them in an osbuild blueprint), and scans for those hardening types **should not** auto-waive failures in those rules. Doing so can lead to hiding bugs.

PR #130, opened 2024-03-26
waive sssd_enable_pam_services because it misses Ansible remediation

vojtapolasek:


mildas:
Unfortunately, it's caused by https://github.com/RHSecurityCompliance/contest/issues/87
It would be better to fix the issue so the contest recognizes such cases.

mildas:
I've updated position of the waiver to have it together with other Ansible waivers. And slightly updated comment. Now lgtm :D

Issue #83, opened 2024-01-11
Support waivers as directories


mildas:
It's complicated to orientate in a waiver file. If you want to add a new waiver, you may wonder where to put it in the file. And with growing number of tests, number of waivers will grow as well.

To organize the waivers, support waiver directories is needed. Then, we can split the existing waiver files to multiple waiver categories and make them easier approachable for others. Initial idea of individual waiver files in the directory is:

- newly discovered and unknown issues
- known issues
- expected failures that can't be fixed

The idea of different waiver targets, `upstream` and `release` waivers, must be preserved. But as directories.

PR #132, opened 2024-03-28
Turn waivers into a directory

comps:
Untested, but the directory-traversal logic seems to work when hand-imported and called from a CLI python interpreter.

It is a Draft because we need to discuss / agree on waiver file names and structure inside `conf/waivers/` here.

I would have liked to keep `CONTEST_WAIVERS` instead of `CONTEST_WAIVER_DIR`, but chances are it would break a lot of things where we have `CONTEST_WAIVERS` pointing to ie. `waivers-upstream`, so the logic would fail to find `$repodir/waivers-upstream/*`.

Fixes https://github.com/RHSecurityCompliance/contest/issues/83 .


mildas:
Directory structure and waiver file names are good. I don't have any objection. Changes lgtm as they are, so if you move it from draft, I'll merge it.

comps:
Tested via the new runcontest, seems to work (assuming runcontest isn't broken).

comps:
I've also moved `file_permissions_unauthorized_world_writable` to `10-unknown`, since it doesn't really have an issue linked, or a satisfactory explanation.

PR #133, opened 2024-04-02
account for [customizations.openscap] being added by oscap to Blueprint

comps:
This makes the test simply modify the pre-existing `[customizations.openscap]` section, rather than add the custom section, enabling compatibility with newer `openscap` versions.

### DIFF
--- a/lib/osbuild.py
+++ b/lib/osbuild.py
@@ -176,6 +176,7 @@ class Blueprint:
                 self.from_oscap(profile)
                 + '\n# ---------- oscap blueprint ends here ----------\n'
             )
+            util.log(f"generated blueprint from oscap:\n{textwrap.indent(self.assembled, '    ')}")
         else:
             self.assembled = f'{self.HEADER}\n\n'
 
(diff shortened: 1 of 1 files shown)


lib/osbuild.py
```diff
@@ -221,14 +222,17 @@ def add_partition(self, mountpoint, minsize):
...
-        self.assembled += util.dedent(fr'''
-            [customizations.openscap]
-            profile_id = "xccdf_org.ssgproject.content_profile_{profile}"
-            datastream = "{ds_file}"
-        ''') + '\n'
+    def set_openscap_datastream(self, ds_file):
+        if '[customizations.openscap]' not in self.assembled:
+            raise SyntaxError("openscap section not found")
+        # flags=re.M seems broken with '^' and '?' together,
+        # it seems to ignore '^' and match like '.*datastream = ...'
```
mildas, line 229: Is it a problem? Regex like `\[customizations\.openscap\]\ndatastream = ...` is fine if you want to `re.M`

comps: Well, the point of `re.M` or `re.MULTILINE` is to make `^` and `$` match every `\n`, not the literal beginning and end of a string, which may contain many `\n` characters.

Funnily enough, I can't seem to reproduce what was causing the issue. I had a situation like
```python
re.sub(r'^#? ?datastream = [^\n]+$', 'replaced', 'blabla\nsome text datastream = foo\nblabla', flags=re.M)
```
and it was replacing it, by `replaced`, where with the manual `\n` and without `re.M` it was fine.

I'll rewrite it back to `re.M` and we'll see. I can't include the section header easily, because `openscap` generates it like
```
[customizations.openscap]
profile_id = "xccdf_org.ssgproject.content_profile_ospp"
# If your hardening data stream is not part of the 'scap-security-guide' package
# provide the absolute path to it (from the root of the image filesystem).
# datastream = "/usr/share/xml/scap/ssg/content/ssg-xxxxx-ds.xml"
```

comps: Although we could just append a line instead of replacing `datastream`, if we want to rely on `openscap` always having it commented out.

mildas: Well, relying on `openscap` having it commented out is fine to me. If appending a line makes it simpler, why not

mildas (approved):

PR #135, opened 2024-04-02
ignore SSL certificate expirations in /static-checks/html-links

comps:
Not 100% sure about this, as we then wouldn't detect links with expirations going on for a long time (effectively rendering the link broken), as opposed to occasional random fails caused by the admin forgetting and fixing it in 1-7 days.

But I think I would rather /not/ see a long-broken link caused by expiration, than observe false positives pop up much more frequently.

Plus, we still do see `warn` results in HTML reports, so chances are that a long-term failing link would be recognized as "hey I've seen that `warn` before", triggering an investigation.

### DIFF
--- a/conf/waivers-upstream
+++ b/conf/waivers-upstream
@@ -269,6 +269,11 @@
 
 # HTML links from datastreams waivers
 #
+# ignore SSL certificate expirations in html-links - these are generally
+# harmless (expiration is not MITM) while being the biggest contributor
+# to false positives, so just ignore them, avoiding frequent random fails
+/static-checks/html-links/.+
+    Match("failed: certificate has expired" in note, sometimes=True)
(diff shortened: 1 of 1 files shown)


conf/waivers-upstream
```diff
@@ -11,6 +11,11 @@
 # because it requires manual configuration which cannot be attained with our rules
 /hardening/.+/service_sssd_enabled
     True
+# ignore SSL certificate expirations in html-links - these are generally
+# harmless (expiration is not MITM) while being the biggest contributor
+# to false positives, so just ignore them, avoiding frequent random fails
+/static-checks/html-links/.+failed: certificate has expired.+
```
mildas, line 17: Move it next to existing html links waiver

PR #150, opened 2024-04-16
Add TMT plans for all our notable use cases

comps:
I split these according to what seemed reasonable. Milan mentioned we should probably use auto-ptp via Beaker directly, so I didn't include a plan for it.

Feel free to add suggestions what to change - I'm not sure if having automated errata runs is even useful if we test everything before that. And we got rid of manual tests since (replaced them with JIRA).

I left the original tags (`NoProductization`, `daily`, etc.) in place, due to TCMS, but I expect we remove them after we switch away - https://github.com/RHSecurityCompliance/contest/issues/151 .

### DIFF
--- a/plans/errata.fmf
+++ b/plans/errata.fmf
@@ -1,0 +1,16 @@
+summary: Testing builds added to errata
+discover:
+    how: fmf
+    filter:
+      - tag:-needs-param
+      - tag:-always-fails
+      - tag:-broken
+    test:
+      # just some basic smoke testing that should never fail
+      - /scanning/oscap-eval
+      - /static-checks
+    exclude:
+      # often fails on temporary retrieval issues
+      - /static-checks/html-links
+
+# vim: syntax=yaml
--- a/plans/stabilization.fmf
+++ b/plans/stabilization.fmf
@@ -1,0 +1,9 @@
+summary: Pre-release "stabilization" testing
+discover:
+    how: fmf
+    # all tests are included by default, except these
+    filter:
+      - tag:-needs-param
+      - tag:-broken
+
+# vim: syntax=yaml
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -68,3 +68,47 @@ they don't install virtual machines.
 
 These are things like `grep`-ing for specific strings (not) present in the built
 content, syntax-checking Ansible playbooks, or verifying HTTP URLs.
+
+# Test tags
+
+These are some of the commonly-used tags amongst tests.
+
+Note that we use tags to indicate properties of tests, not to categorize them
+(think: "needs virtualization", not: "runs during release testing").
+
+## `needs-param`
+
+This indicates a test that is used as a "tool" in automation-assisted use
+cases. It should not run automatically in regular "all tests" runs, as it
+requires the user to give it input via environment variables (parameters).
+
+## `always-fails`
+
+This is a test that uses the `fail` status to indicate some unwanted findings,
+expecting the user to review the list manually. These `fail`s should not be
+waived automatically as they are specific to the configuration the user
+requested.
+
+A test like this is another form of a "tool" and should not be run regularly
+in use cases that expect `pass` to be the norm and `fail` to be a regression.
+
+## `broken`
+
+This is a perfectly valid working test, but the functionality it tests is
+either completely broken, or under very active development, creating interface
+incompatibilities, such as config directive changes, and frequently breaking
+the test.
+
+Despite this, we don't want to disable the test outright, as it is useful for
+debugging and stabilizing the tested functionality via manual use.
+
+However a test like this should not be run by automation, it is not useful
+for preventing regressions.
+
+## `destructive`
+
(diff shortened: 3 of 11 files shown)


plans/errata.fmf
```diff
@@ -0,0 +1,5 @@
+summary: Testing builds added to errata
+discover:
+    how: fmf
+    test:
+      - /scanning/oscap-eval
```
mildas, line 5: Currently, we treat Errata testing as sanity testing `oscap-eval`, `anaconda/ospp`, `ansible/ospp`, (I don't see it in last runt but I'd expect also `oscap/ospp`), `rpmbuild-ctest` and few tests we don't have in contest yet (profile name diff check, no rule removed from benchmark check, container scanning).

It might seem as redundant, but I'd like to stay on the safer side here. Our stabilization build might not be identical to what will be built in downstream and thus few sanity checks seem appropriate.

plans/stabilization.fmf
```diff
@@ -0,0 +1,8 @@
+summary: Pre-release "stabilization" testing
+discover:
+    how: fmf
+    filter:
+      - tag:-adhoc
+      - tag:-broken
+environment:
+    CONTEST_CONTENT_BRANCH: master
```
mildas, line 8: Stabilization is done on stabilization branch. Unfortunately, stabilization branch doesn't have static branch name as it has version suffix `stabilization-v0.1.X`.

plans/errata.fmf
comps: Hmm, so maybe it would make sense to:
1. test as much as possible in stabilization
2. test only a few things in gating CI
3. re-test as much as possible in errata testing (incl. manual?)

?

It shouldn't take that much extra resources to just re-run the automated testing on the final downstream content RPM. Either the full stabilization set (all arches) or just the productization test set (without diffs, x86_64 only).

plans/stabilization.fmf
comps: Right, I'm not even sure if I want to keep `CONTEST_CONTENT_BRANCH: master` here if it should be user-specified on TMT CLI (via runcontest or otherwise). ... And if we can't embed it for stabilization, maybe that's another reason to not have it in the plans.

plans/errata.fmf
mildas: Depends how long it will take to re-test everything on all supported RHELs (usually ~7 versions) and archs in Erratum.
Now it's cheap - you get results fast and it's simple to review. I want to keep it like that. The latter part depends if waivers are in good shape, if so it remains simple. The first part to me is unknown, but I don't want to wait 2 days for results when we switch erratum NEW->QE.
With that, I'd prefer the 2nd option you wrote but slightly adjusted - productization test set with diffs

comps:
We should also consider splitting `productization` into ie. `daily` and `weekly`.

The reason is our ever-growing test coverage - we already have barely enough resources to run tests daily. Once we add UEFI testing next to `with-gui`, it will grow even more. Not to mention the fact that we already don't test non-x86_64 outside of stabilization (presumably).

What will happen when we start auto-scheduling up to 2 reruns of failed/errored tests?

Maybe we should run
* `/plans/daily` - set of tests that can 100% finish overnight (from cca 00:00 to 07:00)
  * all of `/hardening/{oscap,anaconda,image-builder,ansible}`, but without any extra variants (eg. no gui/uefi)
  * one gui test per hardening type (probably STIG)
  * one uefi test per hardening type (maybe OSPP?)
  * x86_64 only, VMs only, no `host-os`
  * anything reasonably cheap from `/scanning` and `/static-checks`
  * on supported RHEL majors (latest minors we plan to target with next SSG release)
* `/plans/weekly` - to be run over weekend (saturday 00:00 until monday 07:00; 24+24+7 hours)
  * everything except any manual tasks (which include `/static-checks/diff` and any other tool-like tests)
  * incl. all GUI/UEFI tests, incl. `host-os` **on all architectures**
  * on supported RHEL majors (latest minors we plan to target with next SSG release)
* `/plans/stabilization`
  * like `weekly`, but incl. manual tests (or tests needing manual review)
  * on multiple ZStreams too, not just one minor per major

(alternatively, we could make `weekly` run `x86_64` only, but on all ZStreams)

comps:
As agreed with @mildas , I have split productization into daily/weekly test sets, for now.

With that, this PR is ready for review.

If merged, I plan to use `/plans/daily` + `/plans/weekly` for `runcontest`.

mildas (commented):
`broken` tag is misleading. You have to read the tag description to understand it's about broken functionality it tests, not broken test. If we could agree on something clearer, it would be great. (I couldn't come with something better during review so I can live with `broken` tag though)

comps:
Rebased, amended.

mildas:
@comps `suspended` doesn't sound much better than `broken`, to that I'd rather stay with `broken`.

`snapshotable` (or just `snapshot`) tag sounds ok

comps:
Ok, changed back to `broken` + left GUI only for `stig_gui`. No need for `snapshottable`, so I didn't add it.

PR #155, opened 2024-04-22
Add `/static-checks/diff` tests, for datastream diffing

comps:
(Go commit-by-commit, there are two somewhat unrelated commits at the start.)

These tests always select old/new datastreams to compare:
* if `CONTEST_CONTENT*` was provided (when using Contest via tmt)
  * use `CONTEST_CONTENT` as new
  * use installed SSG RPM as old
* else (when using Contest via `wow` and `--brew-task`)
  * use installed SSG RPM as new
  * find + download + extract and older SSG RPM version from dnf repositories, use it as old

They always report differences as `fail` results, to show up for human review.

The idea is to run them once, in stabilization, and have a human go through the list of `fail`s in some system that allows commenting on them (TCMS / ReportPortal / etc.). This allows the human to "waive" the failures for that specific stabilization test run, without using static copies of content (a.k.a. "normatives").

As such, these tests should not be run anywhere else, as they would likely always fail and need human intervention. Thus the `NoProductization`.

(The exception is `removed-rules`, which should never fail.)

...
Note that the variable (`refine-value`) diffing groups the results by variable name, as it seemed more natural to have the old/new value for a variable next to each other.  
OTOH rule names are diffed as big `-` / `+` blocks, because it seemed more natural to see what was removed and what was added, rather than interleaving `-` / `+` and sorting it by rule name.

### DIFF
--- a/lib/oscap.py
+++ b/lib/oscap.py
@@ -1,28 +1,161 @@
 import sys
 import re
+import enum
 import contextlib
+import collections
+import types
+import xml.etree.ElementTree as ET
 from pathlib import Path
 
 from lib import util, results
 
-_no_remediation_cache = None
+
+class Datastream:
+    FixType = enum.Flag('FixType', ['bash', 'ansible'])
+
+    def __init__(self, xml_file=None):
+        # extracted datastream metadata
+        #   self.profiles = {
+        #     'ospp': namespace(
+        #       .title = 'Some text',
+        #       .rules = set( 'audit_delete_success' , 'service_firewalld_enabled' , ...),
+        #       .values = set( ('var_rekey_limit_size','1G') , ...),
+        #     ),
+        #   }
+        #   self.rules = {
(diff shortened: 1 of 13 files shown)


lib/oscap.py
```diff
@@ -1,28 +1,144 @@
...
+                parser.feed(chunk)
+                for event, elem in parser.read_events():
+                    if event == 'start':
+                        stack.append(elem)
+                    else:
+                        # optimize a bit - filter out elements too shallow for anything below
+                        if len(stack) < 4:
+                            stack.pop()
+                            continue
+                        frames = [elem.tag.partition('}')[2] or elem.tag for elem in stack]
```
matusmarhefka, line 61: Some comment here describing why we partition using '}' would help the readability, I guess it is because of namespaces right? E.g. `{http://scap.nist.gov/schema/scap/source/1.2}component-ref`.

PR #159, opened 2024-04-25
Remove waivers for firewalld_loopback_traffic rules

marcusburghardt:
Remove the waivers for firewalld_loopback_traffic_trusted rules in Anaconda tests after the
https://github.com/ComplianceAsCode/content/pull/11868

### DIFF
--- a/conf/waivers/20-long-term
+++ b/conf/waivers/20-long-term
@@ -17,9 +17,6 @@
 /hardening/anaconda(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
 # https://github.com/ComplianceAsCode/content/issues/11625
 /hardening/image-builder(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
-# https://github.com/ComplianceAsCode/content/issues/10939
-# https://github.com/ComplianceAsCode/content/issues/10613
-/hardening/anaconda(/with-gui)?/[^/]+/firewalld_loopback_traffic_(restricted|trusted)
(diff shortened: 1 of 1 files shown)


comps:
@marcusburghardt Would that apply to the ImageBuilder waiver too (the issue seems closed as well)?
```
/hardening/image-builder(/with-gui)?/[^/]+/firewalld_loopback_traffic_(restricted|trusted)
```

marcusburghardt:
> @marcusburghardt Would that apply to the ImageBuilder waiver too (the issue seems closed as well)?
> 
> ```
> /hardening/image-builder(/with-gui)?/[^/]+/firewalld_loopback_traffic_(restricted|trusted)
> ```

I doesn't seem the case. At least the image-builder tests are all green, so I removed only the waivers for Anaconda.
Maybe we could keep the image-builder waiver for now and see the results tomorrow.

comps:
> > @marcusburghardt Would that apply to the ImageBuilder waiver too (the issue seems closed as well)?
> > ```
> > /hardening/image-builder(/with-gui)?/[^/]+/firewalld_loopback_traffic_(restricted|trusted)
> > ```
> 
> I doesn't seem the case. At least the image-builder tests are all green, so I removed only the waivers for Anaconda. Maybe we could keep the image-builder waiver for now and see the results tomorrow.

IB tests are green because we ignore any and all failures in them (via other waivers) due to the current instability of oscap/osbuild APIs. So the waiver is likely invalid, but I guess we'll figure out when we re-enable the tests.

PR #164, opened 2024-04-26
Slice per-rule/ list by picking every N-th rule from list

### LINKED ISSUE #163: Randomize rules in `/per-rule`
The number of Automatus (unit) tests varies across rules - some rules have none, some have few fast ones (under 5 minutes), some have lots that take 2 hours to run.

The problem is that these are not randomly distributed, but rather batched together, so that ie. `audit_*` rules get tested by the same parallelized `/per-rule/N` instance, leading to huge swings in elapsed time distribution.

Instead, use a pseudorandom (deterministic) function to distribute the input list of rules across all `/per-rule/N` runners somewhat evenly, applied by each runner before passing the pseudo-randomized list to the slicing function.

mildas:
Fixes #163

### DIFF
--- a/per-rule/test.py
+++ b/per-rule/test.py
@@ -12,25 +12,12 @@ from lib import util, results, oscap, virt, versions, ansible
 from conf import partitions
 
 
-def slice_list(full_list, divident, divisor):
+def slice_list(full_list, start, modulus):
     """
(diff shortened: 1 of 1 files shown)


mildas:
Closing in favor of #166

PR #172, opened 2024-05-01
Add accounts_tmout is a DISA misalignment wavier

Mab879:
See https://github.com/ComplianceAsCode/content/issues/11548


comps:
(push to re-run checks)

PR #176, opened 2024-05-03
waive disa misalignment test involving login_session_timeout rule

vojtapolasek:
the inapplicability of the rule on RHEL 9.0 is deliberate. The underlying feature is not present there. DISA should change their rule or the feature should be ported.


mildas:
@vojtapolasek please rebase and fix conflicts

vojtapolasek:
@mildas  done

PR #178, opened 2024-05-06
Remove sebool_selinuxuser_execmod from the waviers

Mab879:
The latest tests show this passing.

### DIFF
--- a/conf/waivers/10-unknown
+++ b/conf/waivers/10-unknown
@@ -176,5 +176,7 @@
 /hardening/image-builder/anssi_bp28_high/mount_option_tmp_noexec
 /hardening/image-builder/anssi_bp28_high/sebool_polyinstantiation_enabled
     True
+/hardening/image-builder/hipaa/sebool_selinuxuser_execmod
+    rhel == 9
 
 # vim: syntax=python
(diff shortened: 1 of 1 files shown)


comps:
Note that this result is only possible due to our productization upgrading to `openscap-1.3.10` - with old `openscap`, all `/hardening/image-builder` tests error out, and the fail/pass would not appear.

However, looking at both RHEL-9 and RHEL-8 results, the fail still seems to be present on RHEL-9:
```
2024-05-06 04:45:58 test.py:31: lib.results.report_plain:182: WARN sebool_selinuxuser_execmod (waived fail)
```
(as shown in `testout.log` of the RHEL-9 Beaker recipe).

So this waiver should be split off to a separate section.

conf/waivers/10-unknown
```diff
@@ -175,7 +175,6 @@
 /hardening/image-builder/anssi_bp28_high/mount_option_nodev_nonroot_local_partitions
 /hardening/image-builder/anssi_bp28_high/mount_option_tmp_noexec
 /hardening/image-builder/anssi_bp28_high/sebool_polyinstantiation_enabled
-/hardening/image-builder/hipaa/sebool_selinuxuser_execmod
     True
```
comps, line 178:
```suggestion
    True
/hardening/image-builder/hipaa/sebool_selinuxuser_execmod
    rhel == 9
```

PR #181, opened 2024-05-06
Remove error IB waiver.

mildas:
Our tests should use latest available openscap build where blueprint is properly generated with openscap section.

### DIFF
--- a/conf/waivers/20-long-term
+++ b/conf/waivers/20-long-term
@@ -91,11 +91,6 @@
 
 # Image Builder
 #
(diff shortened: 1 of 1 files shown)


comps:
(I'm running all `/hardening/image-builder` for RHEL-8 to double check this doesn't introduce new errors.)

comps:
Looks good, but please modify the `with-gui` waiver right below this one and remove `rhel == 9` from it, so it catches RHEL-8 too, as `with-gui` are expected to error everywhere.

PR #184, opened 2024-05-07
add contest-sshvm, a script for ssh-ing into installed VMs

comps:
This should help a lot with debugging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,12 +157,29 @@ of the modules present in `lib`.
 
 (TODO: probably move to its own document?)
 
+### SSH into Anaconda
+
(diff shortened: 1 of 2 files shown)


hardening/oscap/test.py
```diff
@@ -26,6 +26,7 @@
     g.prepare_for_snapshot()
 
 with g.snapshotted():
+    results.report_and_exit()
```
matusmarhefka, line 29: Why do you call `results.report_and_exit()` here? Maybe leftover from some testing?

comps: Yeah, sorry, I did `git commit -a` :smile:

matusmarhefka (approved):

PR #191, opened 2024-05-14
Ansible playbook syntax check test

mildas:
To check syntax of generated (`oscap xccdf generate fix`) and shipped playbooks.

### DIFF
--- a/lib/util/content.py
+++ b/lib/util/content.py
@@ -57,7 +57,7 @@ def get_playbook(profile):
 
 
 def iter_playbooks():
-    for name in _find_playbooks().iterdir():
+    for name in _find_playbooks().rglob('*'):
         if name.suffix == '.yml':
             yield name
 
--- a/lib/util/content.py
+++ b/lib/util/content.py
@@ -34,22 +34,32 @@ def get_datastream():
     return datastream
 
 
-def get_playbook(profile):
+def _find_playbooks():
     if user_content:
         build_content(user_content)
-        playbook = user_content / 'build' / 'ansible' / f'rhel{rhel.major}-playbook-{profile}.yml'
+        return user_content / 'build' / 'ansible'
     else:
-        base_dir = Path('/usr/share/scap-security-guide/ansible')
-        if rhel.is_true_rhel():
-            playbook = base_dir / f'rhel{rhel.major}-playbook-{profile}.yml'
-        elif rhel.is_centos():
-            if rhel <= 8:
-                playbook = base_dir / f'centos{rhel.major}-playbook-{profile}.yml'
-            else:
-                playbook = base_dir / f'cs{rhel.major}-playbook-{profile}.yml'
+        return Path('/usr/share/scap-security-guide/ansible')
+
+
+def get_playbook(profile):
+    if rhel.is_true_rhel():
+        name = f'rhel{rhel.major}-playbook-{profile}.yml'
+    elif rhel.is_centos():
+        if rhel <= 8:
+            name = f'centos{rhel.major}-playbook-{profile}.yml'
+        else:
+            name = f'cs{rhel.major}-playbook-{profile}.yml'
+    playbook = _find_playbooks() / name
     if not playbook.exists():
         raise RuntimeError(f"cound not find playbook as {playbook}")
     return playbook
+
+
+def iter_playbooks():
+    for name in _find_playbooks().iterdir():
+        if name.suffix == '.yml':
+            yield name
 
 
 def get_kickstart(profile):
--- a/static-checks/ansible/syntax-check/test.py
+++ b/static-checks/ansible/syntax-check/test.py
@@ -1,0 +1,48 @@
+#!/usr/bin/python3
+
+import subprocess
+
+from lib import ansible, util, oscap, results
+
+
+ansible.install_deps()
+
+ds = util.get_datastream()
+all_profiles = oscap.Datastream(ds).profiles
+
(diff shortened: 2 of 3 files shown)


lib/oscap.py
```diff
@@ -37,7 +37,10 @@ def make_rule():
             return types.SimpleNamespace(fixes=self.FixType(0))
         self.rules = collections.defaultdict(make_rule)
         if xml_file:
+            self.path = Path(xml_file)
             self.parse_xml(xml_file)
+        else:
+            self.path = None
```
comps, line 43: This is probably unnecessary, `util.get_datastream()` does the same thing and is more widely used.

I think the mental model around this should be
* `util.get_datastream()` --->  `oscap.global_ds()`

rather than `oscap.global_ds()` being the one-stop-shop for all things DS.

comps:
Regarding shipped playbooks - what about replacing the playbooks-related section of `util/content.py` with something like this (untested)?
```python
def _find_playbooks():
    if user_content:
        build_content(user_content)
        return user_content / 'build' / 'ansible'
    else:
        return Path('/usr/share/scap-security-guide/ansible')


def get_playbook(profile):
    if rhel.is_true_rhel():
        name = f'rhel{rhel.major}-playbook-{profile}.yml'
    elif rhel.is_centos():
        if rhel <= 8:
            name = f'centos{rhel.major}-playbook-{profile}.yml'
        else:
            name = f'cs{rhel.major}-playbook-{profile}.yml'
    playbook = _find_playbooks() / name
    if not playbook.exists():
        raise RuntimeError(f"cound not find playbook as {playbook}")
    return playbook

def iter_playbooks():
    for name in _find_playbooks.iterdir():
        if name.endswith('.yml')
            yield name
```
and then just calling `for playbook in util.iter_playbooks()` ?

static-checks/ansible/syntax-check/test.py
```diff
@@ -0,0 +1,48 @@
...
+from lib import ansible, util, oscap, results
+
+
+ansible.install_deps()
+
+ds = oscap.global_ds()
+all_profiles = ds.profiles.keys()
+
+# Generated playbooks
+for profile in all_profiles:
```
comps, line 14: You should be able to just do
```python
for profile in ds.profiles:
```
as `iter(my_dict)` is the same as `iter(my_dict.keys())`.

static-checks/ansible/syntax-check/test.py
```diff
@@ -0,0 +1,48 @@
...
+    # Get ARF from scan
+    cmd = [
+        'oscap', 'xccdf', 'eval', '--profile', profile,
+        '--results-arf', 'arf.xml', ds.path,
+    ]
+    ret = util.subprocess_run(cmd, stdout=subprocess.DEVNULL)
```
comps, line 20: Why would you suppress the output here and for `generate fix` below? Is it too verbose (100s of KB or MBs)?

Because the output of ie. `oscap` is useful for figuring out whether it froze, and if it isn't excessively large, I would prefer to keep it in the test `output.txt` log.

mildas: Sure, I did suppress it to not polish output because I see that output as not important there as it's about testing playbook syntax.

In `generate fix`, the output is minimal. Some empty lines or so. I'll remove the `stdout=subprocess.DEVNULL` as `generate fix` doesn't polish output much there.
But I do it also for `xccdf eval` few lines above. This one prints all rule results and even with `--progress` it will be quite verbose (scanning with every profile). What about this one? Are you fine with `stdout=subprocess.DEVNULL` there or would you rather see rule results there for figuring out if oscap frozen?

mildas: `generate fix` doesn't output much, but `oscap xccdf eval` outputs all rule results and that's quite verbose even with `--progress` (scan for every profile). Are we fine with that verbose output which isn't specifically relevant for this test case where we test playbook syntax?

comps: I think I would rather have `output.txt` be the ultimate debug log (size within reason, we don't want >1GB logs), so I would say both would be fine. But I don't mind either way, I wast just curious.

lib/util/content.py
```diff
@@ -34,24 +34,38 @@ def get_datastream():
...
+        return user_content / 'build' / 'ansible'
+    else:
+        return Path('/usr/share/scap-security-guide/ansible')
+
+
 def get_playbook(profile):
     if user_content:
         build_content(user_content)
         playbook = user_content / 'build' / 'ansible' / f'rhel{rhel.major}-playbook-{profile}.yml'
     else:
```
comps, line 49: Why did you leave this in the function? ... Isn't it already done by `_find_playbooks()` ?

lib/util/content.py
```diff
@@ -34,24 +34,34 @@ def get_datastream():
...
+    playbook = _find_playbooks() / name
     if not playbook.exists():
         raise RuntimeError(f"cound not find playbook as {playbook}")
     return playbook
 
 
+def iter_playbooks():
+    for name in _find_playbooks().rglob('*'):
+        if name.suffix == '.yml':
+            yield name
```
comps, line 62: I guess this would be "cleaner", as in - use much less resources:
```suggestion
    for dirpath, _, name in _find_playbooks().walk():
        filepath = dirpath / name
        if filepath.suffix == '.yml':
            yield filepath
```
but maybe it doesn't matter much on 1-2 directory levels.

Either would work for me.

PR #196, opened 2024-05-28
Add `/hardening/oscap/old-new` for remediating new content over old one

comps:
This PR
1. moves the "old content" logic from `/static-checks/diff` into `lib.util`, to make it accessible to all tests
2. moves `removed-rules` away from other `diff` tests, as the test doesn't fit the always-fail diffing logic of the other tests, and behaves more like a regular regression test -- this is now possible thanks to the "old content" logic being in `lib.util`
3. adds a `/hardening/oscap/old-new` test that also uses the "old content" logic to
   1. remediate old content
   2. remediate new content
   3. scan new content

The new test is called `old-new` because "package upgrade" would imply it's about upgrading SSG RPM, which it might not be -- the new content would typically come from `CONTEST_CONTENT` or `CONTEST_CONTENT_BRANCH` (from git), upgrading SSG RPM is only one of the tested situations.

Also, I put it under `/hardening` because it reports per-rule results like other hardening tests, unlike some `/scanning` tests. Also to make `/hardening/oscap/.+` waivers applicable to the scan.

`rules_from_verbose()` is a leftover from a previous temporary implementation, but I think it's a good idea to keep it there.

Tested on RHEL-9, tested all tests impacted by this change (`removed-rules`, `diff/*`, `/hardening/oscap/*`) and things seem to be working as expected.


hardening/oscap/old-new/test.py
```diff
@@ -0,0 +1,75 @@
...
+                raise RuntimeError(f"remediation oscap failed with {proc.returncode}")
+            g.soft_reboot()
+
+    # remediate using old content,
+    # then remediate using new content
+    remediate('remediation-old.xml', 'remediation-old.html', 'remediation-old2.html')
+    remediate('remediation-new.xml', 'remediation-new.html', 'remediation-new2.html')
+
+    # scan using new content
+    oval_results = '' if versions.oscap >= 1.3 else '--results results.xml --oval-results'
```
mildas, line 48: Not needed as RHEL8<= has openscap 1.3

mildas (commented):
We don't want it in productization, add `NoProductization` (at least for now, when we still use the tag)

PR #207, opened 2024-06-20
Introduce NIST scapval test

mildas:
NIST scapval tool is external tool used for SCAP content validation. It requires java and our content already have implemented wrapper for running and reporting its results. In this test, both external tool and wrapper are used to check if all our shipped datastreams are SCAP valid.

### DIFF
--- a/lib/util/content.py
+++ b/lib/util/content.py
@@ -12,29 +12,37 @@ if user_content:
     user_content = Path(user_content)
 
 
-def find_datastream_in(root):
-    base_dir = root / Path('usr/share/xml/scap/ssg/content')
-    if rhel.is_true_rhel():
-        return base_dir / f'ssg-rhel{rhel.major}-ds.xml'
-    elif rhel.is_centos():
-        if rhel <= 8:
-            return base_dir / f'ssg-centos{rhel.major}-ds.xml'
-        else:
-            return base_dir / f'ssg-cs{rhel.major}-ds.xml'
-
-
-def get_datastream():
+def _find_datastreams(root):
     if user_content:
         build_content(user_content)
-        datastream = user_content / 'build' / f'ssg-rhel{rhel.major}-ds.xml'
+        datastreams_dir = user_content / 'build'
     else:
-        datastream = find_datastream_in('/')
+        datastreams_dir = root / Path('usr/share/xml/scap/ssg/content')
+    return datastreams_dir
+
+
+def get_datastream(root='/'):
+    if rhel.is_true_rhel():
+        name = f'ssg-rhel{rhel.major}-ds.xml'
+    elif rhel.is_centos():
+        if rhel <= 8:
+            name = f'ssg-centos{rhel.major}-ds.xml'
+        else:
+            name = f'ssg-cs{rhel.major}-ds.xml'
+    datastream = _find_datastreams(root) / name
     if not datastream.exists():
         raise RuntimeError(f"could not find datastream as {datastream}")
     return datastream
 
 
-def _find_playbooks():
+def iter_datastreams(root='/'):
+    for file in _find_datastreams(root).rglob('*'):
+        # Return only DS v1.3, do not return v1.2 (ends with '-ds-1.2.xml')
+        if file.name.endswith('-ds.xml'):
+            yield file
+
+
+def _find_playbooks(root):
     if user_content:
         build_content(user_content)
         return user_content / 'build' / 'ansible'
--- a/lib/util/content.py
+++ b/lib/util/content.py
@@ -47,10 +47,10 @@ def _find_playbooks(root):
         build_content(user_content)
         return user_content / 'build' / 'ansible'
     else:
-        return Path('/usr/share/scap-security-guide/ansible')
+        return root / Path('/usr/share/scap-security-guide/ansible')
 
 
-def get_playbook(profile):
+def get_playbook(profile, root='/'):
     if rhel.is_true_rhel():
         name = f'rhel{rhel.major}-playbook-{profile}.yml'
     elif rhel.is_centos():
@@ -58,16 +58,16 @@ def get_playbook(profile):
             name = f'centos{rhel.major}-playbook-{profile}.yml'
         else:
             name = f'cs{rhel.major}-playbook-{profile}.yml'
-    playbook = _find_playbooks() / name
+    playbook = _find_playbooks(root) / name
     if not playbook.exists():
         raise RuntimeError(f"cound not find playbook as {playbook}")
     return playbook
 
 
-def iter_playbooks():
-    for name in _find_playbooks().rglob('*'):
-        if name.suffix == '.yml':
-            yield name
+def iter_playbooks(root='/'):
+    for file in _find_playbooks(root).rglob('*'):
+        if file.suffix == '.yml':
+            yield file
 
 
 def get_kickstart(profile):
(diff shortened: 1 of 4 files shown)


static-checks/nist-validation/test.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+# Get scapval zip file link from NIST web
+scapval_web = 'https://csrc.nist.gov/Projects/Security-Content-Automation-Protocol/SCAP-Releases/SCAP-1-3' # noqa
+r = requests.get(scapval_web)
+scapval_web_path = re.search(r'href="(.*scapval-.*.zip)"', r.text)
+if not scapval_web_path:
+    raise RuntimeError("SCAPval tool link not found on NIST web")
+scapval_link = 'https://csrc.nist.gov' + scapval_web_path.group(1)
+scapval_zip = scapval_link.split('/')[-1]
```
comps, line 17: Could we please just create a `contest-data` repository here under `RHSecurityCompliance` and store the zip there?

I seem to remember that being the idea when we were discussing it >1 year ago, as there have been a non-trivial amount of failures due to NIST's site being temporarily unreachable at times. Maybe it was due to some throttling of accesses, maybe outages, but I reckon Github has a much higher data hosting capability.

static-checks/nist-validation/test.py
```diff
@@ -0,0 +1,45 @@
...
+
+with util.get_content(build=False) as content_dir:
+    util.subprocess_run(['wget', scapval_link], check=True)
+    util.subprocess_run(['unzip', scapval_zip], check=True)
```
comps, line 21: Doing this in python isn't as difficult as you think:
```python
import io
import zipfile
import requests

url = 'https://github.com/RHSecurityCompliance/contest/archive/refs/heads/main.zip'
reply = requests.get(url)
reply.raise_for_status()
zip = zipfile.ZipFile(io.BytesIO(reply.content))
zip.extractall(path='/tmp/foo')
```
where `/tmp/foo` is `tempfile.TemporaryDirectory()`, or omit the argument altogether if you're comfortable relying on CWD being temp dir.

There are more memory-efficient ways of doing it, like chunking the download via
``` 
with requests.get(url, stream=True) as r:
    r.raise_for_status()
    with open(local_filename, 'wb') as f:
        for chunk in r.iter_content(chunk_size=8192): 
            # write to some Bytes stream from which zipfile can read
```
but 48 MB is probably well within the memory capabilities of any system we run this suite on.

I think I would prefer the short python version over `wget` / `unzip`, but I leave this up to you.

static-checks/nist-validation/test.py
```diff
@@ -0,0 +1,45 @@
...
+
+    ssg_scapval_script = content_dir / 'tests' / 'run_scapval.py'
+    scapval_jar = next(Path.cwd().glob('scapval-*.jar'))
```
comps, line 24: Does it really make sense to use `run_scapval.py` instead of `scapval.sh` directly in the unzipped tool? ... From what I can tell, `run_scapval.py` seems to be mostly about DS version 1.2 vs 1.3, and parsing some results according to some external (XML?) namespace ... ?

Couldn't we just do something like this?
```python
for ds in util.iter_datastreams():
    ds_name = ds.stem
    result_file = f'{ds_name}.result.xml'
    report_file = f'{ds_name}.report.html'
    proc = util.subprocess_run([
        './scapval.sh', .....,
        '-valreportfile', report_file,
        '-file', ds
    ])
    status = 'pass' if proc.returncode == 0 else 'fail'
    results.report(status, name=ds_name, logs=[result_file, report_file])
```
Personally, I think I would prefer reporting each DS from our code, rather than relying on parsing external format. I'm sure this came up in some other test too.

comps (commented):
Added some comments. I'm not going to "force my way" of writing the test, if you decide your original code fits better, I'll respect that.

static-checks/nist-validation/test.py
mildas: Sure, I can do that. However, are we going to check if our repo contains latest scapval? NIST updates scapval from time to time (very rarely). In our previous scapval test, each run we have been comparing our scapval zip SHA with SHA provided by NIST on their web. It's unnecessary complexity I wanted to avoid, thus chosen the path of downloading it directly from web.

static-checks/nist-validation/test.py
comps: Note that I assumed `scapval` would care about exit code, I didn't actually check it. If you end up relying on it, please test it with a malformed XML, to ensure it exits non-0.

static-checks/nist-validation/test.py
comps: Hmm, good point.

On one hand, I would hate either
* having an undocumented process of maybe sometime in the future updating the zip in `contest-data`
* having a documented process of checking for an update, as an extra added complexity for Contest

but I think the real choice is between

1. leaving a potentially old version of scapval in `contest-data`, and updating it when it starts failing (?)
   * without checking any SHA from NIST (would create dependency on NIST being online)
   * this has the advantage of it being under our control, it won't change between stabilization runs during one release, etc.
   * also more reliable hosting
1. always downloading it from NIST

and I think I would still prefer (1), but it's definitely not black-and-white.

I also really don't like the sub-path mangling just to get the zip URL.

lib/util/content.py
```diff
@@ -12,54 +12,62 @@
...
-
-
-def get_datastream():
-    if user_content:
-        build_content(user_content)
-        datastream = user_content / 'build' / f'ssg-rhel{rhel.major}-ds.xml'
-    else:
-        datastream = find_datastream_in('/')
+            name = f'ssg-cs{rhel.major}-ds.xml'
+    datastream = _find_datastreams() / name
```
comps, line 32: This should probably call `_find_datastreams(root)`.

Thinking about it, `_find_datastreams()` should probably have that argument mandatory (without a default), as it's not meant to be used externally. Any default will be sorted out by `get_datastream()` already.

Same for `_find_playbooks()`.

lib/util/old_content.py
```diff
@@ -49,7 +49,7 @@ def _available_ssg_versions():
 def get_old_datastream():
     # installed SSG with datastream in /usr/share/xml
     installed = _installed_ssg_version()
-    root_datastream = util.find_datastream_in('/')
+    root_datastream = util.get_datastream('/')
```
comps, line 52: Honestly, I think calling `get_datastream()` here would be fine. But then `/` is nicely explicit for `root_datastream`. I guess either way works, I just wanted to give you a chance to change it.

static-checks/nist-validation/test.py
```diff
@@ -0,0 +1,36 @@
...
+        '-scapversion', '1.3',
+        '-valreportfile', report_file,
+        '-valresultfile', result_file,
+        '-file', datastream,
+    ]
+    proc = util.subprocess_run(cmd, stdout=subprocess.PIPE, check=True, universal_newlines=True)
+    if 'The target is valid' in proc.stdout:
+        results.report('pass', ds_name)
+    elif 'The target is invalid' in proc.stdout:
+        results.report('fail', ds_name, logs=[report_file, result_file])
```
comps, line 34: Maybe I would add some fall-back in case neither string is present,
```suggestion
        results.report('fail', ds_name, logs=[report_file, result_file])
    else:
        results.report('error', ds_name, "could not parse scapval output")
```
or even a `raise RuntimeError(...)` for terminating the whole test, up to you.

mildas: Updated to `raise RuntimeError`

lib/util/old_content.py
mildas: Yep, I wanted to have `/` explicit, so I kept it like that

comps:
LGTM

PR #208, opened 2024-06-20
enhance docs/WAIVERS.md, describe the waiving algorithm

comps:
This was prompted by more than one person repeatedly misunderstanding the waiving algorithm, so I thought it would be better to just spell it out in the docs.

### DIFF
--- a/docs/WAIVERS.md
+++ b/docs/WAIVERS.md
@@ -19,7 +19,7 @@ By convention, file names inside the waivers directory have specific meaning:
   - note that they may still disappear over time, ie. by dropping support
     for an old OS release, which had the limitation
 
-## Custom file format
+## Sections
 
 The files inside the waivers directory use a custom file format to specify
 when a failure is expected.  
@@ -50,21 +50,6 @@ Each section consists of two parts:
     rhel < 9 and oscap < '1.3.6'
 ```
 
-When a new failing result is to be reported, the sections are evaluated
-sequentially, in order, **from the top**.
-
-Within each section:  
-If at least one of the consecutive regular expressions matches the result name,
-the python expression is evaluated.  
-If that expression also returns `True`, the result is considered waived and
-the traversal stops for that one result (no further sections are tried).
-
-If either the regular expressions don't match, or the python expression returns
-False, the section is skipped and a next one is considered.
-
-If none of the sections match, no waiving happens and the result remains
-unchanged.
-
 ## Regular expressions
 
 The regexps use a pythonic syntax, and are matched using `re.fullmatch()`,
@@ -92,21 +77,7 @@ Or `re.search()` behavior using:
 .*/some/result/name.*
 ```
 
-## Not just a `fail`
-
-The waiving logic is actually evaluated for all of `pass`, `fail` and `error`.
-The `pass` is there to catch any waive matches that suddenly started `pass`-ing,
-and the `error` allows us to waive some infrastructure `error`s as well.
-
-The waiving file format above therefore serves as a generic "matching logic",
-and it's up to the library code to interpret what a match means.
-
-- no match = no change
-- match + pass = fail (unexpected pass)
-- match + fail = warn (waived fail)
-- match + error = warn (waived error)
-
-## Matching and globals
+## Python expressions
 
 The python code block must always be one expression, not a freeform python
 module code.
@@ -122,7 +93,8 @@ The expression has these globals available:
 - `oscap` - an object capable of `openscap-scanner` RPM version comparisons,
   see [versions.oscap](lib/versions.py)
 - `env` - environment variable retrieval function, same as `os.environ.get()`
-- `Match` - a class for complex waive results (see below)
+- `Match` - a class for complex waive results, able to contain both a boolean
+  expression as well as additional parameters
 
 The version comparison objects also support a boolean evaluation, with
 `bool(rhel)` returning `False` if not running on RHEL, and ie. `bool(oscap)`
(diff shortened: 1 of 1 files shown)


docs/WAIVERS.md
```diff
@@ -137,17 +108,87 @@ returning `False` if the RPM is not installed.
...
+This creates a one big unified sequential ordered list of sections
+that is later used for waiving.
 
-```python
-/some/result/name
-    Match(rhel >= 8, strict=True)
+## How a result is waived
+
+When `results.report()` is called by test code to report a new result,
+the waiving logic looks at the status (`pass`, `fail`, etc.).
```
mildas, line 122: Don't mention test implementation details such as `results.report()`. Results reviewer might not be familiar with tests implementations and those lines basically say "see test implementation".

Use rather something like
```suggestion
For each individual result before reporting it, the waiving logic
looks at the status (`pass`, `fail`, etc.).
```
or
```suggestion
Before a result (e.g. rule result or overall test result) is reported,
the waiving logic looks at the status (`pass`, `fail`, etc.).
```

docs/WAIVERS.md
```diff
@@ -137,17 +108,87 @@ returning `False` if the RPM is not installed.
...
+contained within the result, against all regexps in each section.
+
+For example, a test name of `/some/result/name`, when evaluated against
+```
+/different/name.*
+    True
+
+/some/unrelated/string
+/some/.+/name
+    True
```
mildas, line 139: Switch the regexes and use more interesting condition than just `True`

```suggestion
/some/.+/name
/some/unrelated/string
    rhel >= 8
```

It will be useful for further explanation

docs/WAIVERS.md
```diff
@@ -137,17 +108,87 @@ returning `False` if the RPM is not installed.
...
 ```
+would match the second section, and its `/some/.+/name` regexp.
+
+**If no section matches on a regexp, no further processing is done,
+and the result remains intact.**
+
+When a section is matched on regexp, its python expression is evaluated.
+If that expression returns `False`, the section is skipped, and further
+sections (in the big unified list of sections) are tried, to see if
+at least one of them could match.
```
mildas, line 149: Refer the example above here and add extra details
```suggestion
When a section is matched on regexp, its python expression is evaluated.
In our example, `/some/result/name` matched `/some/.+/name` regex. The
python expression to be evaluated for this section is `rhel >= 8`.
If that expression returns `False`, the section is skipped, and result
remains intact for now. Waiving continues with further sections (in the
big unified list of sections) to look for next matching regex.
```

docs/WAIVERS.md
```diff
@@ -137,17 +108,87 @@ returning `False` if the RPM is not installed.
...
+
+**If no section matches on both a regexp AND python expression returning
+True, no further processing is done, the result remains intact.**
+
+When the result matches on both regexp + python of one section, the result
+is subject to waiving, and **no further sections are evaluated for that
+result**, the processing stops on the first matching section.
```
mildas, line 156:
```suggestion
When the result matches on both regexp + python expression, the result
is waived, the processing stops, and waived result (e.g. `warn (waived fail)`)
reported to user. **No further sections are evaluated for that result.**
```

docs/WAIVERS.md
```diff
@@ -137,17 +108,87 @@ returning `False` if the RPM is not installed.
...
+
+If a result got through all that, the waiving logic decides based on the
+reported status (`pass`, `fail`, etc.):
```
mildas, line 159: Too complicated wording, I'd simplify it

```suggestion
If a result got matched by a waiver, the waiving logic decides on reported status:
```

docs/WAIVERS.md
```diff
@@ -137,17 +108,87 @@ returning `False` if the RPM is not installed.
...
+
+- `fail` is changed to `warn`, and "waived fail" is added to `note`
+- `error` is changed to `warn`, and "waived error" is added to `note`
+
+Further, if a waiver returns `strict=True` via `Match()`, or if
+`CONTEST_STRICT_WAIVERS` is set:
```
mildas, line 165: Add extra info about permissive(default) / strict

```suggestion
By default, waivers are permissive causing `pass` result is
reported as `pass (waived pass)` result.

That might be changed to strict via waiver `strict=True`, e.g.
`Match(rhel >= 8, strict=True)`, or running Contest with
`CONTEST_STRICT_WAIVERS`. Adds waiving logic for `pass`:
```

mildas (commented):
Few suggestions

docs/WAIVERS.md
```diff
@@ -137,17 +108,87 @@ returning `False` if the RPM is not installed.
...
+at least one of them could match.
+
+**If no section matches on both a regexp AND python expression returning
+True, no further processing is done, the result remains intact.**
+
+When the result matches on both regexp + python of one section, the result
+is subject to waiving, and **no further sections are evaluated for that
+result**, the processing stops on the first matching section.
+
+If a result got through all that, the waiving logic decides based on the
```
matusmarhefka, line 158:
```suggestion
If a result got through all that and a matching section was found, the waiving logic decides based on the
```

docs/WAIVERS.md
comps: Agree with the non-True condition, but I wanted to emphasize that it's not the first regex that has to match - it can be any of them. And for that, putting it second seemed better.

docs/WAIVERS.md
comps: I think I like my version better - `warn (waived fail)` doesn't appear anywhere (HTML, console, or this document, etc.) and I would really like to emphasize that "the processing stops on the first matching section", to make the reader realize the atomic unit of processing is a section.

docs/WAIVERS.md
comps: Again, I think I like my version better - it doesn't try to make `pass` a special free-form text thing. I'm not sure, "Adds waiving logic for `pass`"  is clearer here.

And `pass (waived pass)` also doesn't appear anywhere.

But I agree that `Match(rhel >= 8, strict=True)` should probably be mentioned, since I removed the section explaining how to use `strict=True`.

docs/WAIVERS.md
mildas: And I'd like to show it doesn't have to be right next to the regex :D What about putting 3 regexes with 2nd as the match?

docs/WAIVERS.md
mildas: You don't have to use exactly what I wrote, we can combine your version with mine. We don't have to mention `warn (waived failed)`. But to me, "the processing stops on the first matching section" says the same as "no further sections are evaluated for that result".

docs/WAIVERS.md
mildas: Sure, you don't have to use exactly what I wrote. But I'd like to have some parts at least clearer.
For example:
```
By default, waivers are permissive causing if `pass` result matches a waiver, the result is intact.
```
or
```
By default, waivers are permissive causing if `pass` result matches a waiver, no `fail`/`error` is reported.
```
"Adds waiving logic for `pass`" can be also changed to something better like "Changes `pass` waiving behavior".

docs/WAIVERS.md
comps: Sure.

docs/WAIVERS.md
comps: I rewrote it slightly, and removed that redundant addition.

matusmarhefka (approved):

PR #211, opened 2024-06-27
Add waivers for recent daily productization findings

mildas:


conf/waivers/10-unknown
```diff
@@ -5,6 +5,10 @@
 # their root cause, filing isseus/bugs or fixing tests as appropriate,
 # eventually either removing the waivers or moving them to other files
 
+# https://github.com/ComplianceAsCode/content/issues/12096
+/hardening/.*/cis.*/sshd_use_approved_ciphers
```
comps, line 9: Let's please calm it down with wildcard matches, I can see this easily matching `cis.*` on a test type instead of a profile name.
```suggestion
/hardening/.*/cis[^/]+/sshd_use_approved_ciphers
```
would at least make sure not further `hier/archy/levels` get matched by the `cis` wildcard.

comps:
Actually, that should have been `[^/]*` because `cis` is also valid.

PR #217, opened 2024-07-02
Move socket_systemd-journal-remote_disabled to long-term waivers.

mildas:
As it has been discovered it's old known OAA issue.

Make it applicable also on other RHEL versions as it's expected to fail everywhere.

### DIFF
--- a/conf/waivers/20-long-term
+++ b/conf/waivers/20-long-term
@@ -51,7 +51,9 @@
 
 # RHEL-8: https://bugzilla.redhat.com/show_bug.cgi?id=1834716
 # RHEL-9: https://bugzilla.redhat.com/show_bug.cgi?id=1999587
+# https://issues.redhat.com/browse/RHEL-45706
 /hardening/anaconda/with-gui/[^/]+/service_avahi-daemon_disabled
+/hardening/anaconda(/with-gui)?/cis[^/]*/socket_systemd-journal-remote_disabled
     True
 
 # https://github.com/ComplianceAsCode/content/issues/11498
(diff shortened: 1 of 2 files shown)


conf/waivers/20-long-term
```diff
@@ -51,7 +51,9 @@
 
 # RHEL-8: https://bugzilla.redhat.com/show_bug.cgi?id=1834716
 # RHEL-9: https://bugzilla.redhat.com/show_bug.cgi?id=1999587
+# https://issues.redhat.com/browse/RHEL-45706
 /hardening/anaconda/with-gui/[^/]+/service_avahi-daemon_disabled
+/hardening/anaconda(/with-gui)?/cis[^/]*/socket_systemd-journal-remote_disabled
     True
```
matusmarhefka, line 57: Does it also happen on RHEL8? Because previously the waiver was RHEL9 only.

conf/waivers/20-long-term
```diff
@@ -51,7 +51,9 @@
 
 # RHEL-8: https://bugzilla.redhat.com/show_bug.cgi?id=1834716
 # RHEL-9: https://bugzilla.redhat.com/show_bug.cgi?id=1999587
+# https://issues.redhat.com/browse/RHEL-45706
 /hardening/anaconda/with-gui/[^/]+/service_avahi-daemon_disabled
+/hardening/anaconda(/with-gui)?/cis[^/]*/socket_systemd-journal-remote_disabled
```
matusmarhefka, line 56: Does it also happen on RHEL8? Because previously the waiver was RHEL9 only.

mildas: It doesn't now, but I expect it could happen also on RHEL8 as the OAA issue is also o RHEL8.

matusmarhefka (approved):

PR #210, opened 2024-06-24
Remove RHEL-7 support

comps:
(And a few related changes.)

PR #223, opened 2024-07-10
Use dynamic ports for BackgroundHTTPServer

comps:
This is a follow-up to https://github.com/RHSecurityCompliance/contest/pull/210/commits/3277a788bb19f9b6b4f1e62665413936d705f255 from https://github.com/RHSecurityCompliance/contest/pull/210.

I've tested
```
/hardening/anaconda/stig$
/hardening/image-builder/stig$
/scanning/disa-alignment/anaconda$
```
(the main/only tests impacted by this) and they all seem to pass/warn just fine, so I presume it works.

I also ran some multi-profile runs of `anaconda` / `image-builder` to make sure there isn't anything persistent that would break 2 or more tests running sequentially on a single system.

### DIFF
--- a/lib/virt.py
+++ b/lib/virt.py
@@ -404,18 +404,11 @@ class Guest:
         if not kickstart:
             kickstart = Kickstart()
 
-        http_port = 8090
-
         kickstart.add_host_repos()
         util.ssh_keygen(self.ssh_keyfile_path)
         with open(f'{self.ssh_keyfile_path}.pub') as f:
             pubkey = f.read().rstrip()
         kickstart.add_authorized_key(pubkey)
-        kickstart.add_install_only_repo(
-            'contest-rpmpack',
-            f'http://{NETWORK_HOST}:{http_port}/repo',
-        )
-        kickstart.add_packages([util.RpmPack.NAME])
 
         disk_extension = 'qcow2' if disk_format == 'qcow2' else 'img'
         disk_path = f'{GUEST_IMG_DIR}/{self.name}.{disk_extension}'
@@ -431,9 +424,18 @@ class Guest:
 
             # host the custom RPM on a HTTP server, as Anaconda needs a YUM repo
             # to pull packages from
-            srv = util.BackgroundHTTPServer(NETWORK_HOST, http_port)
+            srv = util.BackgroundHTTPServer(NETWORK_HOST, 0)
             srv.add_dir(repo, 'repo')
             stack.enter_context(srv)
+
+            # now that we know the address/port of the HTTP server, add it to
+            # the kickstart as well
+            http_host, http_port = srv.server.server_address
+            kickstart.add_install_only_repo(
+                'contest-rpmpack',
+                f'http://{http_host}:{http_port}/repo',
+            )
+            kickstart.add_packages([util.RpmPack.NAME])
 
             ksfile = stack.enter_context(kickstart.to_tmpfile())
 
--- a/lib/virt.py
+++ b/lib/virt.py
@@ -424,13 +424,12 @@ class Guest:
 
             # host the custom RPM on a HTTP server, as Anaconda needs a YUM repo
             # to pull packages from
-            srv = util.BackgroundHTTPServer(NETWORK_HOST, 0)
+            srv = stack.enter_context(util.BackgroundHTTPServer(NETWORK_HOST, 0))
             srv.add_dir(repo, 'repo')
-            stack.enter_context(srv)
+            http_host, http_port = srv.start()
 
             # now that we know the address/port of the HTTP server, add it to
             # the kickstart as well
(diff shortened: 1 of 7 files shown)


comps:
For the record (in case you get the idea during review), this:
```diff
-srv = util.BackgroundHTTPServer(virt.NETWORK_HOST, 0)
-with srv:
+with util.BackgroundHTTPServer(virt.NETWORK_HOST, 0) as srv:
```
doesn't work - the `__enter__` and `__exit__` functions are methods of `srv`, which (the same object) is then further used inside the `with` block. The `as` syntax is for when `__enter__` itself returns something to be used in the block.

I guess we could `return self` from `__enter__`, but maybe that would be even more confusing for readers / maintainers, as there would be 2 not-so-obvious ways to get the object.

lib/virt.py
```diff
@@ -431,10 +424,19 @@ def install(self, location=None, kickstart=None, rpmpack=None, disk_format='raw'
 
             # host the custom RPM on a HTTP server, as Anaconda needs a YUM repo
             # to pull packages from
-            srv = util.BackgroundHTTPServer(NETWORK_HOST, http_port)
+            srv = util.BackgroundHTTPServer(NETWORK_HOST, 0)
             srv.add_dir(repo, 'repo')
             stack.enter_context(srv)
```
mildas, line 429: To be consistent with other `BackgroundHTTPServer` use cases, shouldn't it be like this?

```suggestion
            srv = util.BackgroundHTTPServer(NETWORK_HOST, 0)
            stack.enter_context(srv)
            srv.add_dir(repo, 'repo')
```

mildas (commented):
> I guess we could return self from __enter__, but maybe that would be even more confusing for readers / maintainers, as there would be 2 not-so-obvious ways to get the object.

What do you mean by `there would be 2 not-so-obvious ways to get the object`? If we return `self` from `__enter__` and use it only in `with`+`as`, then there would be basically "only" one way we would be okay with.

lib/virt.py
comps: No, it's better to add the file before starting the server if possible (to fully eliminate any chance of race conditions).

In the other cases, it's not possible.

comps:
> > I guess we could return self from **enter**, but maybe that would be even more confusing for readers / maintainers, as there would be 2 not-so-obvious ways to get the object.
> 
> What do you mean by `there would be 2 not-so-obvious ways to get the object`? If we return `self` from `__enter__` and use it only in `with`+`as`, then there would be basically "only" one way we would be okay with.

The problem is the `enter_context()` cases where we need to

1. get the server object, for adding files to
2. add files to the object
3. start the object as a server

with some other http-server-unrelated code in between.

It's a subtle difference but we would essentially be doing the equivalent of
```python
with ExitStack() as stack:
    f = open('file.txt')
    stack.enter_context(f)
```
which is not the same as
```python
with ExitStack() as stack:
    stack.enter_context(open('file.txt'))
```
due to how the context manager is called.

---

Maybe I should side-step this issue and a `.start()` function, to remove the confusion of the cases mentioned above,
```python
with util.BackgroundHTTPServer('127.0.0.1', 0) as srv:
    srv.add_file(...)
    srv.start()   # can be swapped with add_file()
    ...
```
or
```python
srv = util.BackgroundHTTPServer('127.0.0.1', 0)
srv.add_file(...)
stack.enter_context(srv.start())   # can be swapped with add_file()
```
?

lib/virt.py
mildas: Why not? For example in `hardening/anaconda` you can add `remediation-ds.xml` outside of `srv` context and so `add_file` before context as well.

comps: Hmm, that's a good point - the anaconda use case could probably be done before starting the server.

I'll fix it and add `.start()` to the server, so that object initialization is more clearly decoupled from starting the server and we don't rely on the sneaky context manager behavior.

comps:
I've pushed an alternate approach - have a look at the module help for `lib/util/httpsrv.py`.

Hopefully this makes things a bit clearer.

comps:
Tested on `-t '/hardening/image-builder/(ospp|stig)$' -t '/hardening/anaconda/(ospp|stig)$' -t '/scanning/disa-alignment/anaconda$' --arch x86_64 --rhel 9 --rhel 8` and it seems to work.

mildas:
Yep, code looks good

Nit-pick: f8af92cb95adb3f800d0816f6bae37b1bade89e4 is unnecessary now. Could you drop it?

comps:
> Yep, code looks good
> 
> Nit-pick: [f8af92c](https://github.com/RHSecurityCompliance/contest/commit/f8af92cb95adb3f800d0816f6bae37b1bade89e4) is unnecessary now. Could you drop it?

Done.

PR #224, opened 2024-07-11
Fix rpmbuild-ctest to use also upstream Content

mildas:
Use `util.get_content` to get SSG content. If content is provided by env variable, use it, not the packaged one. Otherwise, use packaged content. In both cases, built it and test it using `cmake`.

### DIFF
--- a/static-checks/rpmbuild-ctest/main.fmf
+++ b/static-checks/rpmbuild-ctest/main.fmf
@@ -6,6 +6,8 @@ environment+:
 duration: 1h
 require+:
   - rpm-build
+  - cmake
+  - python3-jinja2
 recommend+:
   - dnf-utils
   - yum-utils
--- a/static-checks/rpmbuild-ctest/main.fmf
+++ b/static-checks/rpmbuild-ctest/main.fmf
@@ -8,10 +8,13 @@ require+:
   - rpm-build
   - cmake
   - python3-jinja2
+  - python3-devel
+  - python3-pip
 recommend+:
   - dnf-utils
   - yum-utils
   - yum-builddep
+  - bats
 tag:
   - CI-Tier-1
   - Errata
--- a/static-checks/rpmbuild-ctest/test.py
+++ b/static-checks/rpmbuild-ctest/test.py
@@ -29,6 +29,10 @@ else:
         '-DSSG_BUILD_SCAP_12_DS=OFF',
     ]
 
+# Extra modules to enable more unit tests
+python_modules = ['lxml', 'pytest', 'trestle', 'openpyxl', 'pandas', 'cmakelint']
+util.subprocess_run(['python3', '-m', 'pip', 'install', *python_modules])
+
 with util.get_content(build=False) as content_dir:
     build_dir = content_dir / 'build'
 
--- a/static-checks/rpmbuild-ctest/test.py
+++ b/static-checks/rpmbuild-ctest/test.py
@@ -1,52 +1,55 @@
 #!/usr/bin/python3
 
 import re
-from pathlib import Path
 
-from lib import util, results, dnf
-
-
-def build_source(src_rpm, path=None):
-    """
-    Build package from src rpm.
-
-    'src_rpm' is path to src rpm file.
-    Define 'path' to specify path where to build. Otherwise, build
-    is done to rpmbuild directory in current working directory.
-    """
-    util.subprocess_run(['dnf', 'builddep', '-y', src_rpm], check=True)
-
-    path = Path(path) if path else Path.cwd() / 'rpmbuild'
-
-    util.subprocess_run(['rpm', '-ivh', '--define', f'_topdir {path}', src_rpm], check=True)
-
-    spec_path = next((path / 'SPECS').glob('*.spec'))
-    util.subprocess_run(['rpmbuild', '-bc', '--define', f'_topdir {path}', spec_path], check=True)
-
-    return path
+from lib import util, results, versions
 
 
-# Get src rpm and build from it
-with dnf.download_rpm('scap-security-guide', source=True) as src_rpm:
-    rpm_build = build_source(src_rpm)
-ssg_build = next((rpm_build / 'BUILD').glob('*/build'))
+# options are taken from scap-security-guide spec file
+if versions.rhel.is_true_rhel():
+    cmake_options = [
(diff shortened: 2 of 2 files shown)


static-checks/rpmbuild-ctest/test.py
```diff
@@ -1,52 +1,43 @@
...
+with util.get_content(build=False) as content_dir:
+    build_dir = content_dir / 'build'
+
+    # no default product, no derivatives, build only one specific rhel
+    cmd = [
+        'cmake', '../',
+        f'-DSSG_PRODUCT_RHEL{versions.rhel.major}:BOOLEAN=TRUE',
+        '-DSSG_PRODUCT_DEFAULT:BOOLEAN=FALSE',
+        '-DSSG_SCIENTIFIC_LINUX_DERIVATIVES_ENABLED:BOOL=OFF',
+        '-DSSG_CENTOS_DERIVATIVES_ENABLED:BOOL=OFF',
```
comps, line 17: Does this really work with CentOS Stream (in CaC/content Github CI) ?  
(`SSG_PRODUCT_RHEL9.5` + `CENTOS_DERIVATIVES_ENABLED` OFF)

static-checks/rpmbuild-ctest/test.py
```diff
@@ -1,52 +1,43 @@
...
+        '-DSSG_ANSIBLE_PLAYBOOKS_PER_RULE_ENABLED:BOOL=ON',
+    ]
+    util.subprocess_run(cmd, cwd=build_dir)
+    cmd = ['make', '-j4']
+    util.subprocess_run(cmd, cwd=build_dir)
+    # ctest
```
comps, line 23: Let's use some vertical space between these commands + let's not define `cmd` for one short command which can be inlined:
```suggestion

    util.subprocess_run(['make', '-j4'], cwd=build_dir)
    
    # ctest
```

static-checks/rpmbuild-ctest/test.py
```diff
@@ -1,52 +1,43 @@
...
+    util.subprocess_run(cmd, cwd=build_dir)
+
+    with open(build_dir / 'ctest_results') as f:
+        # Result format: X/Y Test  #X: test_name .................  test_result   Z sec
+        result_regex = re.compile(r'\d+\s+Test\s+#\d+:\s+([^\s]+)\s+\.+\s+(\w+)\s+')
+        for line in f:
+            result_match = result_regex.search(line)
+            if result_match:
+                test_name = result_match.group(1)
+                if result_match.group(2) == 'Passed':
```
comps, line 37:
```suggestion
                test_name, test_result = result_match.groups()
                if test_result == 'Passed':
```
?

static-checks/rpmbuild-ctest/test.py
```diff
@@ -1,52 +1,43 @@
...
+                    result = 'pass'
+                else:
+                    result = 'fail'
```
comps, line 40:
```suggestion
                test_name, test_result = result_match.groups()
                result = 'pass' if test_result == 'Passed' else 'fail'
```
?

static-checks/rpmbuild-ctest/test.py
```diff
@@ -1,52 +1,55 @@
...
+        '-DSSG_SEPARATE_SCAP_FILES_ENABLED=OFF',
+        '-DSSG_BASH_SCRIPTS_ENABLED=OFF',
+        '-DSSG_BUILD_SCAP_12_DS=OFF',
+    ]
+
+with util.get_content(build=False) as content_dir:
+    build_dir = content_dir / 'build'
+
+    util.subprocess_run(['cmake', '../', *cmake_options], cwd=build_dir)
+    util.subprocess_run(['make', '-j4'], cwd=build_dir)
```
comps, line 36: These should also probably have `check=True`, which is not default.

static-checks/rpmbuild-ctest/test.py
```diff
@@ -1,52 +1,55 @@
...
+
+    # ctest
+    cmd = [
+        'cmake', '--build', build_dir, '--target', 'test',
+        '--', 'ARGS=--output-on-failure --output-log ctest_results'
+    ]
+    util.subprocess_run(cmd, cwd=build_dir)
```
comps, line 43: Same here, `check=True`.

mildas: Not sure, I'll have to check it

mildas: Return code 2 when any of unit tests fails, thus leaving this one without `check=True`

mildas:
@comps besides of added `check=True`, added also more `recommend` packages to enable more ssg unit tests

mildas:
Added new commit to fix search for failed ctests.

Previously, failed and passed results had same format:
```
40/52 Test #40: cce_avail_check-sle15 ............................................   Passed    0.04 sec
41/52 Test #41: utils-build_control_from_reference_sanity ........................   Failed   11.75 sec
```
but now there is:
```
40/52 Test #40: cce_avail_check-sle15 ............................................   Passed    0.04 sec
41/52 Test #41: utils-build_control_from_reference_sanity ........................***Failed   11.75 sec
```
and the regex didn't account for that as it expected spaces after `....`.

Thus, I've changed that to match only the part before result:
`40/52 Test #40: cce_avail_check-sle15 ............................................`
and not to match result. Just find the line and then search for `Passed` in it.

mildas:
@comps There are new packages added in `main.fmf` and some python3 modules are installed via `pip3` in test code. I want to enable as much as possible and as those are not available via dnf, I've chosen `pip3`-way. And intentionally without `check=True` in case any of the modules would be unavailable.

Tested on rhel8 and rhel9, passes. Thus, ready for merge.

static-checks/rpmbuild-ctest/main.fmf
```diff
@@ -6,10 +6,16 @@ environment+:
 duration: 1h
 require+:
```
comps, line 7: You are also missing `python3-pip`.

comps: Actually, in RHEL, `python3` depends on `python3-pip`, so we're good, .... but non-default python versions don't, ie. `python3.12` installs by default without `python3.12-pip`, so maybe I would still add it to keep track of that dependency explicitly.

Your choice.

PR #233, opened 2024-07-18
Initial changes to enable RHEL-10

comps:
See the individual commits - most of them have extra descriptions explaining the changes.

### DIFF
--- a/conf/partitions.py
+++ b/conf/partitions.py
@@ -11,7 +11,7 @@ partitions = [
     ('/boot', 700),
     ('/', 1000),
     ('/home', 100),
-    ('/var', 6000),
+    ('/var', 5000),
     ('/var/log', 1000),
     ('/var/log/audit', 1000),
     ('/var/tmp', 1000),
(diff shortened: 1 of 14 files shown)


comps:
I've added https://github.com/RHSecurityCompliance/contest/pull/233/commits/717c52486c57c59d2343b74d0ac4fa3111a1b219 (as a normal push of an extra commit) and removed (via force push) any commits that were about temporary test disablement - 9d702878f4c1855810, 99c11f4c0f52 and 0048085f4898.

The reason for these is that I was originally hoping to get fully working RHEL-10 testing, with oscap scan results for all tests - sure, there would be failures and/or errors, but waivers could be created in a next step and we could have included RHEL-10 in productization testing.

Unfortunately, both RHEL-10 and CaC/content for it are currently very broken, and achieving that state will require slow and methodical approach of running Contest tests one-by-one and fixing CaC/content and RHEL along the way.  
And for this use case, having all tests enabled (no artificial unselects), and available for content debugging, is probably better.

mildas (approved):

PR #239, opened 2024-07-24
Add `/hardening/kickstart` tests for oscap-generated kickstarts

comps:
(Probably best to go commit-by-commit - the early ones are about cleanup.)

The image-builder (osbuild) rework is mainly to unify the API with `/hardening/kickstart`. I would have done the same for `/hardening/anaconda`, but (IIRC) openscap `%addon` cannot read `file://` URLs or point to local files on the installed OS, it can only reliably use remote HTTP URLs.

### DIFF
--- a/conf/waivers/20-long-term
+++ b/conf/waivers/20-long-term
@@ -9,6 +9,8 @@
 # which are not available in their final form in the Anaconda environment
 # - see https://github.com/ComplianceAsCode/content/issues/9746
 /hardening/anaconda(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
+# - possibly unrelated https://github.com/ComplianceAsCode/content/issues/12276
+/hardening/kickstart(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
 # https://github.com/ComplianceAsCode/content/issues/11625
 /hardening/image-builder/[^/]+/firewalld_sshd_port_enabled
     True
@@ -17,6 +19,9 @@
 # and a later enable_authselect remediation breaks it
 # - see https://github.com/OpenSCAP/openscap/issues/1880
 /hardening/anaconda/.+/accounts_password_pam_retry
+# - happens for kickstarts too, even when those should not fail on ordering,
+#   https://github.com/ComplianceAsCode/content/issues/12277
+/hardening/kickstart/.+/accounts_password_pam_retry
     True
 
 # caused by one of:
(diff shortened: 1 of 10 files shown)


comps:
Added a condition to exclude RHEL-8 after a brief message with Evgeny - we probably want to support/test RHEL-9 in addition to RHEL-10.

One last step is to add waivers.

comps:
I didn't join the `/hardening/anaconda` waivers with `/hardening/kickstart` because
* most of the kickstart waivers have one common CaC/content issue files
* they have a chance of being fixed, rather than abandoned as with OAA
* they (as a group) probably share a common problem, and I didn't want to fragment that over multiple sections like with `/hardening/anaconda`

mildas (commented):
`kickstart` part is fine.
For `image-builder` move, why not to move also rule unselect to test code? Unselect rules in `contest-ds.xml` and add that modified datastream to `rpmpack`. No need for oscap tailoring blueprint section. That will unify unselect part with OAA/kickstart where unselect is done directly in datastream without providing tailoring file.

comps:
Changed `/hardening/image-builder` to be more like the kickstart test,
```diff
--- hardening/kickstart/test.py 2024-08-08 21:21:32.803772075 +0200
+++ hardening/image-builder/test.py     2024-08-15 13:38:57.474565165 +0200
@@ -3,3 +3 @@
-import os
-
-from lib import util, results, virt, oscap
+from lib import results, oscap, osbuild, util
@@ -9 +7 @@
-virt.Host.setup()
+osbuild.Host.setup()
@@ -11 +9 @@
-g = virt.Guest()
+g = osbuild.Guest()
@@ -18 +16 @@
-# tell the 'oscap xccdf eval --remediate' in %post to use it
+# tell the 'oscap xccdf eval --remediate' called by osbuild-composer to use it
@@ -24 +22 @@
-    'fix', '--fix-type', 'kickstart',
+    'fix', '--fix-type', 'blueprint',
@@ -28,4 +26 @@
-ks = virt.translate_oscap_kickstart(lines, '/root/remediation-ds.xml')
-
-if os.environ.get('USE_SERVER_WITH_GUI'):
-    ks.packages.append('@Server with GUI')
+blueprint = osbuild.translate_oscap_blueprint(lines, '/root/remediation-ds.xml')
@@ -33 +28 @@
-g.install(kickstart=ks, rpmpack=rpmpack)
+g.create(blueprint=blueprint, rpmpack=rpmpack)
@@ -52 +47 @@
-results.report_and_exit(logs=['report.html', 'results-arf.xml.gz'])
+results.report_and_exit(logs=['report.html', 'results-arf.xml.gz', g.osbuild_log])
```

mildas (approved):

PR #242, opened 2024-07-29
Introduce `subset-profile` tag

### LINKED ISSUE #229: Test only the largest superset of rules for a given profile type
An idea by @mildas - we could significantly reduce our testing matrix if we tested only
* `anssi_bp28_enhanced`
  * instead of all of `enhanced` , `high` , `intermediary` , `minimal`
* `ccn_advanced`
  * instead of all of `advanced` , `intermediate`, `minimal`
* maybe something with CIS and (not) testing all levels and server/workstation ?

It isn't 100% coverage as the profiles might theoretically have different variables, so a profile with the largest set of rules might still behave differently within the subset of rules of a smaller profile, ... but I think that's a reasonable trade-off for the significant testing matrix reduction.

mildas:
`subset-profile` tag is for profiles that are subset of other profile. Those rulesets are covered by the other profile (superset) but variables might differ and thus it's good to test those from time to time.

In this PR, couple of "new" profiles (subset profiles) are introduced. Subset profiles are tagged and using the tag excluded from daily productization as we don't want to test all those profiles every day.

Fixes #229

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -112,3 +112,16 @@ unusable for further testing, typically by hardening it.
 
 A test that just installs extra RPMs from the package manager, or enables
 extra services, is not considered destructive.
+
+## `subset-profile`
+
+Some compliance standards specify multiple "levels" of system hardening,
+with the more permissive levels typically being a subset of the more
+restrictive ones, ie. Minimal, Intermediary and Enhanced.
+
+This tag signifies that a test covers one of the levels (profiles), which
+are themselves a subset of another level (profile).
+
+There are other minor factors (content profile variables) that break this
+relationship, however, if you want a smaller test set at the cost of ignoring
+these factors, exclude tests with this tag.
--- a/conf/waivers/10-unknown
+++ b/conf/waivers/10-unknown
@@ -164,8 +164,8 @@
 # Image Builder
 #
 # TODO: file issues ?
-/hardening/image-builder/anssi_bp28_high/mount_option_tmp_noexec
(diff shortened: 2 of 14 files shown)


docs/TESTS.md
```diff
@@ -112,3 +112,10 @@ unusable for further testing, typically by hardening it.
 
 A test that just installs extra RPMs from the package manager, or enables
 extra services, is not considered destructive.
+
+## `subset-profile`
+
+A tests for a profile which is a subset of other profile test.
+
+A subset profile's ruleset is covered by the superset. However, profile
+variables in the superset profile might be different.
```
comps, line 121: I would personally like a bit more background for this. Feel free to copy/paste from this:

```
Some compliance standards specify multiple "levels" of system hardening,
with the more permissive levels typically being a subset of the more
restrictive ones, ie. Minimal, Intermediary and Enhanced.

This tag signifies that a test covers one of the levels (profiles), which
are themselves a subset of another level (profile).

There are other minor factors (content profile variables) that break this
relationship, however, if you want a smaller test set at the cost of ignoring
these factors, use this tag.
```
It would personally help me when I'm staring at this in 4 years, wondering why the tag exists. :smile:

comps (changes requested):

mildas:
I'd add that I haven't tagged `e8` profile with the tag. `ism_o` profile is a superset of `e8` in RHEL8 and RHEL9, however not in RHEL10. I've opened this topic with developers that we should align it. But for now, I leave it as not tagged and we will see how which way content goes.

PR #244, opened 2024-07-29
extend waiver for timer_dnf-automatic_enabled to RHEL 9

vojtapolasek:
Related issue: https://github.com/ComplianceAsCode/content/issues/12234


comps:
The condition probably should have been `or` instead of `and`?

AFAIK no RHEL version is both 8 and 9. :)

PR #251, opened 2024-08-06
Fix "missing tar command" error in /per-rule tests

mildas:
`/per-rule` tests were not executed as they were terminated at the beginning because of
```
ERROR - Terminating due to error: Unable to upload test scripts: Cannot extract data tarball /root/ssgts/tmppt219c40.tar.gz.
```
After checking logs, it was caused by missing `tar` package which is not installed on minimal RHEL9 by default.

### DIFF
--- a/hardening/anaconda/test.py
+++ b/hardening/anaconda/test.py
@@ -16,7 +16,7 @@ profile = util.get_test_name().rpartition('/')[2]
 ks = virt.translate_ssg_kickstart(profile)
 
 if os.environ.get('USE_SERVER_WITH_GUI'):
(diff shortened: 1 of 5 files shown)


comps:
Hmm, there is `add_packages()` of `class Kickstart` rather than adding it directly, but our guidelines (https://github.com/RHSecurityCompliance/contest/blob/main/docs/STYLE.md#avoid-getterssetters) say to use it the way you did in this PR.

I'm guessing this was mainly due to `add_package_group()` being different on RHEL-7.

Would you mind replacing all the uses of `add_packages()` and `add_package_group()` with simple `.append()` in other tests, as a separate commit?

PR #255, opened 2024-08-16
Misc RHEL-10 fixes, `rhc-worker-playbook`, OSPP disablement

comps:
The OSPP change is a result of https://github.com/ComplianceAsCode/content/pull/12223.

### DIFF
--- a/hardening/anaconda/main.fmf
+++ b/hardening/anaconda/main.fmf
@@ -66,6 +66,10 @@ adjust:
 /ism_o:
 
 /ospp:
+    adjust+:
+      - when: distro >= rhel-10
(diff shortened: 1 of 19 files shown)


comps:
Regarding the disablement of OSPP and the need to change so many files -- unfortunately, there is not reasonable solution for us for now, we really do need https://github.com/teemtee/tmt/issues/1793 .

We could do ie. `PROFILE=blabla /hardening/oscap`, but that specification can't be easily paralellized via runcontest, for example, because it doesn't represent FMF tests, only env+test combos, and (even then) the test specification would break some assumptions about test/result name prefixing.

PR #262, opened 2024-09-17
remove no longer valid waiver for audit_rules_privileged_commands

vojtapolasek:
The issue has been closed and I think the solution was to move whole Audit group out of the System group. This allows better ordering.


comps:
LGTM, the Github checks are failing because of broken Ubuntu, but the change looks innocent enough, I'll try to merge it

PR #265, opened 2024-09-25
Introduce test for checking rule identifiers

mildas:
Check for known references of individual profiles.

References are links with specific identifiers.
```
<xccdf-1.2:reference href="https://www.cisecurity.org/benchmark/red_hat_linux/">2.3.3</xccdf-1.2:reference>
```
It's not possible to automatically check whether the identifiers are correct as those are coming from policy requirements. However, we can test if policy references are assigned to profile rules.

The references mapping could be read from CaC/content `ssg/constants.py`. But as it lists references for all CaC/content profiles and products, most of that stuff would have to be filtered out and wouldn't simplify things much here. And so mapping is part of the test code.

~~Note: do not merge yet, a lot of failures caused by missing references in various profiles. I'm going to create CaC upstream issues for each individual rhel+profile combination and create waivers.~~

### DIFF
--- a/static-checks/rule-identifiers/main.fmf
+++ b/static-checks/rule-identifiers/main.fmf
@@ -1,0 +1,10 @@
+summary: Search for policy references in rules
+test: python3 -m lib.runtest ./test.py
+result: custom
+environment+:
+    PYTHONPATH: ../..
+duration: 10m
+adjust:
+  - when: distro == rhel-10
+    enabled: false
+    because: rhel10 profiles are in draft state with missing references
--- a/static-checks/rule-identifiers/main.fmf
+++ b/static-checks/rule-identifiers/main.fmf
@@ -4,7 +4,3 @@ result: custom
 environment+:
     PYTHONPATH: ../..
 duration: 10m
-adjust:
-  - when: distro == rhel-10
-    enabled: false
-    because: rhel10 profiles are in draft state with missing references
--- a/static-checks/rule-identifiers/test.py
+++ b/static-checks/rule-identifiers/test.py
@@ -1,0 +1,52 @@
+#!/usr/bin/python3
+
+import re
+from collections import defaultdict
+
+from lib import util, results, oscap
+
+
+profile_references = {
+    # srg, disa, stigid@PRODUCT or controls file
+    'stig': [
+        'https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cgeneral-purpose-os',  # noqa:E501
+        'https://public.cyber.mil/stigs/cci/',
+        'https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux',
+    ],
+    # ospp
+    'ospp': ['https://www.niap-ccevs.org/Profile/PP.cfm'],
+    # cis@PRODUCT or controls file
+    'cis': ['https://www.cisecurity.org/benchmark/red_hat_linux/'],
+    # anssi (controls file)
+    'anssi': ['https://cyber.gouv.fr/sites/default/files/document/linux_configuration-en-v2.pdf'],
+    # ism
+    'ism_o': ['https://www.cyber.gov.au/acsc/view-all-content/ism'],
+    # hipaa
+    'hipaa': ['https://www.gpo.gov/fdsys/pkg/CFR-2007-title45-vol1/pdf/CFR-2007-title45-vol1-chapA-subchapC.pdf'],  # noqa: E501
+    # pcidss4
+    'pci-dss': ['https://docs-prv.pcisecuritystandards.org/PCI%20DSS/Standard/PCI-DSS-v4_0.pdf'],
+}
+
+
+profiles = oscap.global_ds().profiles
+# Parse from datastream all rules and all their references
+rule_references = defaultdict(set)
+for frames, elements in oscap.parse_xml(util.get_datastream()):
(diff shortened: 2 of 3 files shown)


static-checks/rule-identifiers/main.fmf
```diff
@@ -0,0 +1,10 @@
+summary: Search for policy references in rules
+test: python3 -m lib.runtest ./test.py
+result: custom
+environment+:
+    PYTHONPATH: ../..
+duration: 10m
+adjust:
+  - when: distro == rhel-10
+    enabled: false
+    because: rhel10 profiles are in draft state with missing references
```
comps, line 10: Wouldn't waivers be better here, to not forget about this as RHEL-10 develops further?

static-checks/rule-identifiers/test.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/python3
+
+import re
+from collections import defaultdict
+
+from lib import util, results, oscap, versions
+
+
+profile_references = {
```
comps, line 9: Consider moving this to a separate file, so it's easier to navigate/edit for developers (potentially). But it's not a big deal, keeping it in the test might keep the information fragmentation low.

static-checks/rule-identifiers/test.py
```diff
@@ -0,0 +1,63 @@
...
+    'cis': ['https://www.cisecurity.org/benchmark/red_hat_linux/'],
+    # anssi (controls file)
+    'anssi': ['https://cyber.gouv.fr/sites/default/files/document/linux_configuration-en-v2.pdf'],
+    # ism
+    'ism_o': ['https://www.cyber.gov.au/acsc/view-all-content/ism'],
+    # hipaa
+    'hipaa': ['https://www.gpo.gov/fdsys/pkg/CFR-2007-title45-vol1/pdf/CFR-2007-title45-vol1-chapA-subchapC.pdf'],  # noqa: E501
+}
+# Version specific references
+if versions.rhel.major == 8:
```
comps, line 28: I don't know where this originated - I've seen other Contest PRs with `versions.rhel.major`, but IMHO we shouldn't use it for comparisons like this.

I would prefer
```suggestion
if versions.rhel == 8:
```

static-checks/rule-identifiers/test.py
```diff
@@ -0,0 +1,63 @@
...
+    # pcidss
+    profile_references['pci-dss'] = ['https://www.pcisecuritystandards.org/documents/PCI_DSS_v3-2-1.pdf']  # noqa: E501
+elif versions.rhel.major == 9:
+    # pcidss
+    profile_references['pci-dss'] = ['https://www.pcisecuritystandards.org/documents/PCI_DSS_v4_0.pdf']  # noqa: E501
+
+
+# Parse from datastream all rules and all their references
+def get_rule_references():
```
comps, line 37: You already call the function on the module level and do it only once. I would get rid of it and just put the code into the test directly.

static-checks/rule-identifiers/test.py
```diff
@@ -0,0 +1,63 @@
...
+        rule, reference = elements[-2:]
+        # TODO: use str.removeprefix on python 3.9+
+        rule = re.sub('^xccdf_org.ssgproject.content_rule_', '', rule.get('id'))
+        reference = reference.get('href')
+        references[rule].add(reference)
+
+    return references
+
+
+profiles = oscap.Datastream(util.get_datastream()).profiles
```
comps, line 52: No need to re-parse the XML, you can get this from the already-parsed DS.
```suggestion
profiles = oscap.global_ds().profiles
```
Using `oscap.Datastream()` was really meant only for tests that operate on more than one datastream (such as diff/comparison tests).

static-checks/rule-identifiers/test.py
```diff
@@ -0,0 +1,63 @@
...
+rule_references = get_rule_references()
+
+for profile, references in profile_references.items():
+    for rule in profiles[profile].rules:
+        for reference in references:
+            if reference in rule_references[rule]:
+                results.report('pass', f'{rule} ({profile})')
```
comps, line 59: I'm not sure if we want `(` and `)` and spaces in the result (test) name. YAML supports it, our waivers would also deal with it (spaces in the regexp should be fine), but it would be better to avoid it.

Maybe something like
```suggestion
                results.report('pass', f'{profile}/{rule}')
```
?

static-checks/rule-identifiers/test.py
```diff
@@ -0,0 +1,63 @@
...
+            else:
+                results.report('fail', f'{rule} ({profile})', f'missing {reference} ({profile})')
```
comps, line 61: Same as above, plus the fact that mentioning profile again is probably just noise in the note (it's already part of the result name).
```suggestion
                results.report('fail', f'{profile}/{rule}', f"missing {reference}")
```

comps (changes requested):

static-checks/rule-identifiers/test.py
mildas: I prefer having it here as you mentioned to keep information fragmentation low.

static-checks/rule-identifiers/main.fmf
mildas: Honestly, not sure if waivers would be better as it doesn't change much in forgetting about it. (supposing our current usual workflow - waiver, upstream issue, and forget).

Still I'll think about it. But as RHEL10 are still deeply in draft state, those waiver+upstream issue would be without `productization-issue` label and thus easily lost in issue backlog.

PR #272, opened 2024-10-08
increase disk size of installed guests to 30 Gb

vojtapolasek:
Recently, it seems that current 20 Gb is not enough.
I recently tried some Anaconda installation with my Autocontest project (simple interface to this project) and it seems that 30 Gb  is curently fine.


vojtapolasek:
Closing, it seems the problem has been solved rather by https://github.com/RHSecurityCompliance/contest/pull/273

PR #275, opened 2024-10-18
Update nist-validation test to provide more granular results

matusmarhefka:
Scapval tool is failing when we build SCE content by default in RHEL 9 and RHEL 10 data streams because it doesn't expect content to use checking systems other than the OVAL and OCIL (base requirement `SRC-118`). For more details see https://github.com/ComplianceAsCode/content/pull/12488

We can waive this fail, it shouldn't cause any problems for 3rd party scanners as our content still contains also OVAL checks. To do so the test has been updated to parse XML results file generated by the scapval tool.

The test is also updated to work on RHEL 10 where `java-21-openjdk` is the default as scapval tool has no problem running with this newer version of java.

### DIFF
--- a/static-checks/nist-validation/test.py
+++ b/static-checks/nist-validation/test.py
@@ -5,6 +5,7 @@ import io
 import zipfile
 import requests
 import subprocess
+import xml.etree.ElementTree as ET
 
 from lib import util, results
 
@@ -15,6 +16,8 @@ r.raise_for_status()
 zip = zipfile.ZipFile(io.BytesIO(r.content))
 zip.extractall()
 os.chmod('scapval.sh', 0o755)
+
+ns = {'nist': 'http://csrc.nist.gov/ns/decima/results/1.0'}
 
 for datastream in util.iter_datastreams():
     ds_name = datastream.stem
@@ -27,12 +30,19 @@ for datastream in util.iter_datastreams():
         '-valresultfile', result_file,
         '-file', datastream,
     ]
-    proc = util.subprocess_run(cmd, stdout=subprocess.PIPE, check=True, universal_newlines=True)
-    if 'The target is valid' in proc.stdout:
-        results.report('pass', ds_name)
-    elif 'The target is invalid' in proc.stdout:
-        results.report('fail', ds_name, logs=[report_file, result_file])
-    else:
-        raise RuntimeError("SCAPval out has not been correctly parsed")
+    util.subprocess_run(cmd, stdout=subprocess.DEVNULL, check=True)
+    tree = ET.parse(result_file)
+    root = tree.getroot()
+    for elem in root.findall('./nist:results/nist:base-requirement', ns):
+        name = f'{ds_name}/{elem.attrib["id"]}'
+        status = elem.find('./nist:status', ns).text
+        if status in ['NOT_TESTED', 'NOT_APPLICABLE']:
+            continue
+        elif status in ['PASS', 'WARNING', 'INFORMATIONAL']:
+            results.report('pass', name)
+        elif status == 'FAIL':
+            results.report('fail', name, logs=[report_file, result_file])
+        else:
+            results.report('error', name, logs=[report_file, result_file])
 
 results.report_and_exit()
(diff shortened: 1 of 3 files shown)


static-checks/nist-validation/test.py
```diff
@@ -27,12 +30,15 @@
...
+    util.subprocess_run(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
+    tree = ET.parse(result_file)
+    root = tree.getroot()
+    failures = 0
+    for elem in root.findall('./nist:results/nist:base-requirement/[nist:status="FAIL"]', ns):
+        name = f'{ds_name}/{elem.attrib["id"]}'
+        results.report('fail', name, logs=[report_file, result_file])
+        failures += 1
+    if not failures:
         results.report('pass', ds_name)
```
mildas, line 42: Let's search all requirements' statuses, report `fail` for `status="FAIL"`, `pass` for PASS, and `error` (?) for everything else. To have everything reported and to avoid the `failures` counted.

static-checks/nist-validation/test.py
```diff
@@ -27,12 +30,15 @@
         '-valresultfile', result_file,
         '-file', datastream,
     ]
-    proc = util.subprocess_run(cmd, stdout=subprocess.PIPE, check=True, universal_newlines=True)
-    if 'The target is valid' in proc.stdout:
+    util.subprocess_run(cmd, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, check=True)
```
comps, line 33: Why mute stderr? What if the tool fails for some unknown reason and we won't be able to debug it?

Does it produce too much output on stderr?

jan-cerny: normally it doesn't produce anything on stderr

but it produces a lot on stdout

static-checks/nist-validation/test.py
```diff
@@ -27,12 +30,15 @@
...
+    tree = ET.parse(result_file)
+    root = tree.getroot()
+    failures = 0
+    for elem in root.findall('./nist:results/nist:base-requirement/[nist:status="FAIL"]', ns):
+        name = f'{ds_name}/{elem.attrib["id"]}'
+        results.report('fail', name, logs=[report_file, result_file])
+        failures += 1
+    if not failures:
```
jan-cerny, line 41: I guess `failures` can be bool?

static-checks/nist-validation/test.py
matusmarhefka: @mildas Based on the scapval stdout where it summarizes results there are the following result types:
```
57 base requirements having 48 PASS, 6 WARNING, 1 FAIL, and 2 INFORMATIONAL results
```
Why do you want to report other results than FAIL? It will increase the amount of results in the report which I don't think we will be looking into.

static-checks/nist-validation/test.py
matusmarhefka: Good point, I will update it to still print stderr.

static-checks/nist-validation/test.py
matusmarhefka: Okay, let me update it then.

static-checks/nist-validation/test.py
```diff
@@ -27,12 +30,17 @@
...
+    root = tree.getroot()
+    for elem in root.findall('./nist:results/nist:base-requirement', ns):
+        name = f'{ds_name}/{elem.attrib["id"]}'
+        status = elem.find('./nist:status', ns).text
+        if status in ['NOT_TESTED', 'NOT_APPLICABLE']:
+            continue
+        elif status in ['PASS', 'WARNING', 'INFORMATIONAL']:
+            results.report('pass', name)
+        else:
+            results.report('fail', name, logs=[report_file, result_file])
```
mildas, line 44: Maybe?
```suggestion
        elif status in ['PASS', 'WARNING', 'INFORMATIONAL']:
            results.report('pass', name)
        elif status == 'FAIL':
            results.report('fail', name, logs=[report_file, result_file])
         else:
            results.report('error', name, logs=[report_file, result_file])
```

Issue #19, opened 2023-05-25
Test on UEFI/non-UEFI


comps:
Currently, the virtual machines used by `lib/virt.py` are using the traditional "BIOS" scheme. This unfortunately doesn't exercise all of the content rules - there are several variants on uefi/non-uefi bootloader ones, for example.

Libvirt/QEMU can create UEFI (and even Secure Boot capable) virtual machines, but not on RHEL-7. So investigate this after we drop RHEL-7 support, and add UEFI testing to our testing matrix. Either as default + BIOS testing as a special case, or vice versa.

I imagine it like `with-gui` tests, but consider reducing the scope by adding it to `oscap` only (not `anaconda`, `ansible`, etc.) to save execution time. Or not, do whatever, I'm not your father.

comps:
This is now even more important for RHEL-10.

PR #276, opened 2024-10-18
Add `/hardening/*/uefi` tests

comps:
This fixes https://github.com/RHSecurityCompliance/contest/issues/19 .

These tests are especially important for RHEL-10 (which might be UEFI-by-default), but they seem to work well on others, so I left them without a `distro` limitation. https://github.com/ComplianceAsCode/content/issues/12510 and https://github.com/ComplianceAsCode/content/issues/12508 were already found using them (on non-RHEL-10).

Yes, they increase the testing matrix a bit, but I would argue these are just as useful as `with-gui` variants, if not more.

Regarding missing `/hardening/anaconda`, the commit message explains it:
```
    /hardening/anaconda is not included because the upstream kickstarts
    are not UEFI compatible, ie.

        Failed to find a suitable stage1 device: EFI System Partition
        cannot be of type xfs.; EFI System Partition must be mounted on
        one of /boot/efi.; EFI System Partition cannot be of type lvmpv.
```

---

I'd like to merge these as `broken` for now (not run in any automation), because while they **do** work, it's usually only on the 3rd retry on RHEL-9.5, something is causing the VM to fail to start and/or get DHCP, with about 70% reproducibility rate.

Having these tests merged would help long-term debugging of the issue (so I didn't want to just delete my work), and I'll create a GH issue to make sure we don't forget, but I also feel uncomfortable adding more broken-ish tests to our regular pipelines, we already have `image-builder` for that .. :grin:

### DIFF
--- a/conf/waivers/20-long-term
+++ b/conf/waivers/20-long-term
@@ -12,7 +12,7 @@
 # - possibly unrelated https://github.com/ComplianceAsCode/content/issues/12276
 /hardening/kickstart(/with-gui)?/[^/]+/firewalld_sshd_port_enabled
 # https://github.com/ComplianceAsCode/content/issues/11625
-/hardening/image-builder/[^/]+/firewalld_sshd_port_enabled
+/hardening/image-builder/.+/firewalld_sshd_port_enabled
     True
 
 # rule ordering issue - accounts_password_pam_retry is checked first and passes,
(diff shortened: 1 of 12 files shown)


conf/waivers/20-long-term
```diff
@@ -134,4 +134,13 @@
 /static-checks/rule-identifiers/ism_o/.*
     rhel == 8 or rhel == 9 or rhel == 10
 
+# UEFI/SecureBoot
+#
+# https://github.com/ComplianceAsCode/content/issues/12508
+/hardening/ansible/uefi/anssi_bp28_(enhanced|high)
+    status == 'error'
```
mildas, line 141: Not relevant anymore? https://github.com/ComplianceAsCode/content/pull/12514

mildas (approved):

comps:
> To keep `anaconda` tc consistent with others, what about adding `uefi.fmf` there as well but disabled with reasoning pointing to Content issue regarding kickstart not compatible with uefi?

I think that would make sense on the profile level, but probably not on the test type / variant level. Ie. we already do not have disabled `with-gui` tests for `image-builder`, because it's not meant to be used with GUI, apparently.

comps:
I'll launch a full daily run on this, to make sure it didn't break anything.

PR #277, opened 2024-10-22
Make nist-validation test waiver applicable also on CentOS

matusmarhefka:


conf/waivers/30-permanent
```diff
@@ -89,6 +89,6 @@
 # Caused by SCE content being built by default, enabled
 # in https://github.com/ComplianceAsCode/content/pull/12488
 /static-checks/nist-validation/ssg-rhel9-ds/SRC-118
-    rhel >= 9
+    (rhel.is_centos() and rhel >= 9) or rhel >= 9
```
mildas, line 92: The problem is in regex, not in condition. The string to match on CentOS is
```
/static-checks/nist-validation/ssg-cs9-ds/SRC-118 
```

mildas (changes requested):

PR #278, opened 2024-10-24
Add bootc tests

comps:
The last commit is still WIP and will change, but all commits under it should be ready for review.

### DIFF
--- a/conf/remediation.py
+++ b/conf/remediation.py
@@ -39,4 +39,18 @@ def excludes():
                 'enable_fips_mode',
             ]
 
+    # RHEL Image Mode
+    # TODO: revisit these, see which ones we really need for use with Contest
+    if test_name.startswith('/hardening/container'):
+        rules += [
+            'no_direct_root_logins',
+            'firewalld_sshd_disabled',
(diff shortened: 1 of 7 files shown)


comps:
I've redesigned the bootc-image-builder test quite a bit; turns out we don't really need a `Blueprint` (`config.toml`) since we can do all the configuration in a `Containerfile`, incl. ssh keys for testing. And the `bootc-image-builder` container seems to build `qcow2` just fine without `config.toml`, so it must be just optional.

comps:
I'll be submitting the `anaconda-ostree` test as disabled-by-default because of https://issues.redhat.com/browse/RHEL-66155 :
```
17:20:46,800 INF program: Running in chroot '/mnt/sysroot'... ostree admin instutil set-kargs crashkernel=1G-4G:192M,4G-64G:256M,64G-:512M root=UUID=8cbbd836-5803-4046-93f3-50e7ac200564 rw
17:20:46,830 INF program: error: Unable to find a deployment in sysroot
```

comps:
This should be now ready for review.

matusmarhefka (approved):

PR #279, opened 2024-10-31
Allow also `posix` and `community.general` Ansible modules

mildas:
Modules with these FQCN should be accepted as well, not just modules from `ansible.builtin`.

### DIFF
--- a/static-checks/ansible/allowed-modules/test.py
+++ b/static-checks/ansible/allowed-modules/test.py
@@ -55,6 +55,8 @@ def process_task(task, all_allowed_modules):
             found_allowed_modules.update(process_task(block_task, all_allowed_modules))
     else:
         keywords = set(kw.replace('ansible.builtin.', '') for kw in original_keywords)
+        keywords = set(kw.replace('community.general.', '') for kw in keywords)
+        keywords = set(kw.replace('ansible.posix.', '') for kw in keywords)
         allowed_module = keywords.intersection(all_allowed_modules)
         if allowed_module:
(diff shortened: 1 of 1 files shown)


static-checks/ansible/allowed-modules/test.py
```diff
@@ -55,6 +55,8 @@ def process_task(task, all_allowed_modules):
             found_allowed_modules.update(process_task(block_task, all_allowed_modules))
     else:
         keywords = set(kw.replace('ansible.builtin.', '') for kw in original_keywords)
+        keywords = set(kw.replace('community.general.', '') for kw in keywords)
```
matusmarhefka, line 58: What about something like this:
```
keywords = set()
for kw in original_keywords:
    keywords.add(re.sub(r'(ansible.builtin.|community.general.|ansible.posix.)', r'', kw))
```

mildas: Changes 3 similar lines to 3 different lines and that's fine. But adds complexity and reduces readability. So I'd stay with my solution.

comps: It's probably pointless now, but maybe this would have been a reasonable alternative for avoiding the multiple iterations of `keywords`:
```python
filters = lambda kw: (kw
    .replace('ansible.builtin.', '')
    .replace('community.general.', '')
    .replace('ansible.posix.', '')
)
keywords = set(map(filters, original_keywords))
```
(or a more traditional version with backslashes instead of wrapping the lambda body in `( )`).

PR #286, opened 2024-11-12
Report waived results as pass for TF environment

mildas:
For TF runs, tmt correctly shows waived results as `warn`, but then TF shows them as failures and that a huge mess mainly for our Errata test run where we check for missing identifiers (hundreds of waived rules https://github.com/RHSecurityCompliance/contest/blob/main/conf/waivers/20-long-term#L124).

### DIFF
--- a/lib/waive.py
+++ b/lib/waive.py
@@ -221,6 +221,10 @@ def match_result(status, name, note):
 
 
 def rewrite_result(status, name, note, new_status='warn'):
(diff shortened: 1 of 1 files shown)


comps:
Let's please make a new function, something like `on_testingfarm()`, right below `def have_beaker_api()`, that would
```python
return os.environ.get('TESTING_FARM_REQUEST_ID') is not None
```
and then use that function on the two places.

PR #296, opened 2024-12-20
Add AlmaLinux to SC-118 wavier

Mab879:
Need so https://github.com/ComplianceAsCode/content/pull/12757 can be merged.

### DIFF
--- a/conf/waivers/30-permanent
+++ b/conf/waivers/30-permanent
@@ -97,7 +97,7 @@
 #
 # Caused by SCE content being built by default, enabled
 # in https://github.com/ComplianceAsCode/content/pull/12488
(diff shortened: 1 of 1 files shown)


comps:
@Mab879 What's the solution here? ... Is it possible to configure CaC/content to *not* run Contest on anything except RHEL / CentOS Stream? Thanks!

Mab879:
> @Mab879 What's the solution here? ... Is it possible to configure CaC/content to _not_ run Contest on anything except RHEL / CentOS Stream? Thanks!

Current no. Since in that PR Alma is set as a derivative RHEL it builds and works like CentOS Stream. Converting to draft until we know what approach we taking in the above mentioned PR.

Mab879:
No longer needed.

PR #297, opened 2025-01-02
Waive #12766

jan-cerny:


conf/waivers/10-unknown
```diff
@@ -5,6 +5,14 @@
 # their root cause, filing isseus/bugs or fixing tests as appropriate,
 # eventually either removing the waivers or moving them to other files
 
+# https://github.com/ComplianceAsCode/content/issues/12766
```
matusmarhefka, line 8: You are missing a condition after these waivers, see https://github.com/RHSecurityCompliance/contest/blob/main/docs/WAIVERS.md

Also in https://github.com/ComplianceAsCode/content/issues/12766 you mention some failures are only on RHEL 10 and some are on RHEL <= 9 so we should mark the respective waivers based on that.

jan-cerny:
I have add the conditions

jan-cerny:
I have improved the conditions

PR #307, opened 2025-01-16
Report additional Automatus-generated log files in `/per-rule` tests

comps:
These might be helpful for debugging.


comps:
Seems to work on RHEL-8 now too.

matusmarhefka (approved):

PR #308, opened 2025-01-17
Waive bootc stig test on RHEL 10

jan-cerny:
The fail of /hardening/container/bootc-image-builder/stig is caused by https://issues.redhat.com/browse/RHEL-74244.


jan-cerny:
We have discussed it with @matusmarhefka  and we think It would be better to not waive it so that the fail is visible because it's a blocking issue.

PR #309, opened 2025-01-22
Use bootc-ready OpenSCAP from Copr/Packit only on <= RHEL-9

comps:
(Also waive a few RHEL-10 related issues.)

### DIFF
--- a/conf/waivers/20-long-term
+++ b/conf/waivers/20-long-term
@@ -145,4 +145,11 @@
 /hardening/image-builder/uefi/.+/mount_option_boot_efi_nosuid
     True
 
+# Container / Image Mode remediations
+#
+# enable_authselect failing on image-mode
+# https://issues.redhat.com/browse/OPENSCAP-5272
+/hardening/container/bootc-image-builder/[^/]+/enable_authselect
+    rhel == 9
+
 # vim: syntax=python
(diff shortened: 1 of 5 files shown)


conf/waivers/20-long-term
```diff
@@ -145,4 +145,11 @@
 /hardening/image-builder/uefi/.+/mount_option_boot_efi_nosuid
     True
 
+# Container / Image Mode remediations
+#
+# enable_authselect failing on image-mode
+# https://issues.redhat.com/browse/OPENSCAP-5272
+/hardening/container/bootc-image-builder/[^/]+/enable_authselect
+    True
```
matusmarhefka, line 153: It's issue only on RHEL9 so let's go with `rhel == 9`.

hardening/container/bootc-image-builder/main.fmf
```diff
@@ -25,7 +25,7 @@ adjust:
   - when: arch != x86_64
     enabled: false
     because: we want to run virtualization on x86_64 only
-  - when: distro < rhel-9.6
+  - when: distro < rhel-9 or distro ~< rhel-9.6
```
matusmarhefka, line 28: But this will allow the test to run also on RHEL8 which doesn't make sense.

matusmarhefka (approved):

PR #334, opened 2025-02-10
Order `sshd.service` after `network-online.target` to scan a fully booted OS

comps:
This is a series of 3 commits that logically follow each other, to implement a fix for https://github.com/ComplianceAsCode/content/issues/12870 and https://github.com/ComplianceAsCode/content/issues/12831 .

From the commit descriptions:

### `add RpmPack to all our tests, always`
```
This allows us to customize the testing environment anywhere.
```

### `add file contents support for RpmPack`
```
This allows adding arbitrary text contents as a file path on the target
where the RPM is to be installed.
```

### `hack sshd.service to always start after network-online.target`
```
This fixes a rare (but unfortunate) issue where some systemd unit files
(ie. timers like dnf-automatic.timer) depend on network-online.target
and are thus started later than sshd with its network.target.

In rare cases, this leads to us connecting to the system and scanning
it sooner than all services/timers can fully start up.

Since this is somewhat unrealistic scenario, it would make sense to
wait with scanning a while longer, simulating a "fully booted" OS.
To achieve that, the (probably) easiest and most predictable way
is to just make sshd.service start *after* network-online.target,
letting our wait-for-ssh logic simly wait a few seconds longer.
```


comps:
This is now ready for merging.

jan-cerny:
I have tried to run this using autocontest.sh. This doesn't fix the issue with dnf-automatic.timer (https://github.com/ComplianceAsCode/content/issues/12831). In my example, in the journalctl log I can see that the oscap scan started earlier than dnf-automatic.timer. The rule timer_dnf-automatic_enabled was evaluated 4 seconds before the dnf-automatic.timer started. But, I'm not sure if I run the item correctly, so I will retry tomorrow and will let you know.

comps:
@jan-cerny Can you check `systemctl show sshd.service` and see if `network-online.target` is on the `After=` line ?

(IOW if the fix was applied at least.)

comps:
I think I figured out what is wrong - see the attached `systemd-analyze` graph. It's `multi-user.target` starting much later (10s after `network-online.target`).

I'll rework the PR to make sshd start after `multi-user.target` instead.

plot

comps:
I've re-done the 3-4 commits to use `multi-user.target`, but also to reload systemd when the service is modified, so `After=` in `systemctl show` is accurate.

And it does seem to work now (see second graph, attached) - I noticed that ssh became available ~10 seconds after the login prompt on a serial console.

Looking at the graph, I suspect timers starting late are due to `chronyd` taking ~10 seconds to synchronize NTP time, and systemd probably wants timers to be accurate, so it waits with their start for that.

plot2

jan-cerny:
After the latest change it started to pass.

I have run this command:

```
./autocontest.sh test -t '/hardening/host-os/ansible/anssi_bp28_minimal' -h hostname   --content-path /home/jcerny/work/git/scap-security-guide/ --contest-path /home/jcerny/work/git/contest/
```

I the journalctl log I can see that the `dnf-automatic.timer` started 6 seconds before the `oscap` evaluation started. In the HTML report the rule `rule_timer_dnf-automatic_enabled` is "pass".

jan-cerny:
> @jan-cerny Can you check systemctl show sshd.service and see if network-online.target is on the After= line ?

No, it isn't

```
[root@scap-t5 ~]# systemctl show sshd.service | grep After=
After=system.slice sysinit.target sshd-keygen.target systemd-journald.socket basic.target network.target
```

jan-cerny:
please resolve the conflicts

comps:
Rebased, conflicts resolved, since @matusmarhefka ACKed this before, I'm merging it now to prevent further conflicts with new work.

PR #343, opened 2025-02-16
Move building of content to python code

comps:
This effectively reverts https://github.com/RHSecurityCompliance/contest/pull/260/commits/9c3184d8717e5e5395f8ddfffdaf31115d3b0617 and replaces it with a more comprehensive tunable python building logic.

The main reason is that some tests already do that (namely `rpmbuild-ctest`) because of specific CMake build options. And I would like to add more - a new `/per-rule` rewrite done via https://github.com/ComplianceAsCode/content/pull/13029 needing `SSG_THIN_DS:BOOL=ON` along with `SSG_BUILT_TESTS_ENABLED:BOOL=ON`.

So via this PR I propose a two-part logic:
* Have tests explicitly request CMake build options if they need them, otherwise default to some defaults
* Pre-build content in runcontest/productization using 16 CPU cores, with some maximum possible option set, then upload built content to all Beaker VMs

All of this is safely guarded by the built options check that re-builds the content if the requested options differ from how it was built:
* Tests that don't care simply don't specify such options (ie. most tests won't care about thin datastreams)
* Tests that care and want them disabled will rebuild content (request option value different)
* Tests that care and want them enabled will also rebuild content (request option value different)

The end result shouldn't have any negative effect on our productization testing (we're not adding build time to test `duration` because we'll pre-build content on Gitlab runner), but should be cleaner for ad-hoc executions since a developer `rsync`ing built content won't accidentally forget `--playbooks-per-rule` (breaking a test) - the test will just rebuild if needed.

I also took the opportunity to document the complexity of `lib.util.content`, and moved building to a separate function, since most `get_content()` users did not want it built anyway.

### DIFF
--- a/lib/util/content.py
+++ b/lib/util/content.py
@@ -1,15 +1,65 @@
+"""
+Utility functions for acquiring binary/source CaC/content.
+
+Content can come from two sources:
+ - CONTEST_CONTENT as a git-repo-style directory
+ - scap-security-guide RPM
+
+Each of these has binary/sources on different places:
+ - CONTEST_CONTENT
+   - binary content needs to be built, is located in build/*
+   - source content already exists as the directory itself
+ - scap-security-guide RPM
+   - binary content already exists in /usr/share
+   - source content needs to be downloaded as SRPM
+
+Unifying these under one API is therefore a significant challenge,
+especially since the scap-security-guide RPM fragments binary content
+to multiple paths in /usr/share.
+
+This module therefore provides
+ - several get_*() functions for getting binary content
+   - if CONTEST_CONTENT is used, the content is built as-needed
+ - one get_content_source() function for getting source content
+   - if scap-security-guide RPM is used, a SRPM gets downloaded + extracted
+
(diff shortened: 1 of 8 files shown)


lib/util/content.py
```diff
@@ -1,15 +1,65 @@
...
+
+
+def get_user_content(build=True):
+    user_content = os.environ.get('CONTEST_CONTENT')
+    if not user_content:
+        return None
+    user_content = Path(user_content).absolute()
+    # variable defined, but path specified does not exist
+    if not user_content.exists():
+        raise RuntimeError("CONTEST_CONTENT={user_content} does not exist")
```
matusmarhefka, line 59: This should be f-string.

lib/util/content.py
```diff
@@ -1,15 +1,65 @@
+"""
+Utility functions for acquiring binary/source OpenSCAP content.
```
matusmarhefka, line 2:
```suggestion
Utility functions for acquiring binary/source CaC/content.
```

matusmarhefka (approved):

PR #346, opened 2025-02-20
OPENSCAP-5414 Change DISA Alignment test

jan-cerny:
The DISA Alignment test had problem that it doesn't account for situations when a single SSG rule maps to multiple DISA rules or when a single DISA rule maps to multiple SSG rules. In these situations we get false positive results. This commit attempts to fix these issues. The test will not use `compare_results.py` from ComplianceAsCode because that script is too complex but we will process the results by functions in `scanning/disa-alignment/shared.py`.

Fixes: https://github.com/ComplianceAsCode/content/issues/13020

We will also remove the waivers of the upstream issue.

### DIFF
--- a/scanning/disa-alignment/shared.py
+++ b/scanning/disa-alignment/shared.py
@@ -153,7 +153,7 @@ def compare_results(ssg_results, disa_results):
 
 
 def get_disa_result_to_str(stig_ids_results):
-    if stig_ids_results:
+    if len(stig_ids_results) == 0:
         return "Empty"
     ll = (f"{k}:{v}" for k, v in stig_ids_results.items())
     return ", ".join(ll)
--- a/scanning/disa-alignment/shared.py
+++ b/scanning/disa-alignment/shared.py
@@ -1,11 +1,46 @@
-import re
+from collections import namedtuple
+import xml.etree.ElementTree as ET
 
 from lib import results
 
+nsmap = {
+    "xccdf": "http://checklists.nist.gov/xccdf/1.2",
+}
+STIG_REF = "https://public.cyber.mil/stigs/srg-stig-tools/"
+CCE = "https://ncp.nist.gov/cce"
+SSG_RULE_PREFIX = "xccdf_org.ssgproject.content_rule_"
+DISA_RULE_PREFIX = "xccdf_mil.disa.stig_rule_"
+
+DISARuleResult = namedtuple('DISARuleResult', ['rule_id', 'result'])
 
 profile = 'stig'
 
 shared_cmd = ['oscap', 'xccdf', 'eval', '--progress']
+
+class SSGRuleResult:
+    def __init__(self, rule_id, cce_id, rule_title, stig_ids, result):
+        self.rule_id = rule_id
+        self.cce_id = cce_id
+        self.rule_title = rule_title
+        self.stig_ids = stig_ids
+        self.result = result
+        self.stig_ids_results = dict()
+        self.final_result = None
+
+    def __repr__(self):
+        return (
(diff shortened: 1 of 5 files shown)


scanning/disa-alignment/shared.py
```diff
@@ -1,12 +1,47 @@
-import re
+from collections import namedtuple
```
comps, line 1: Per https://github.com/RHSecurityCompliance/contest/blob/main/docs/STYLE.md#import-namespaces , this should probably just `import collections`.
```suggestion
import collections
```

scanning/disa-alignment/shared.py
```diff
@@ -1,12 +1,47 @@
...
 
+nsmap = {
+    "xccdf": "http://checklists.nist.gov/xccdf/1.2",
+}
+STIG_REF = "https://public.cyber.mil/stigs/srg-stig-tools/"
+CCE = "https://ncp.nist.gov/cce"
+SSG_RULE_PREFIX = "xccdf_org.ssgproject.content_rule_"
+DISA_RULE_PREFIX = "xccdf_mil.disa.stig_rule_"
+
+DISARuleResult = namedtuple('DISARuleResult', ['rule_id', 'result'])
```
comps, line 14:
```suggestion
DISARuleResult = collections.namedtuple('DISARuleResult', ['rule_id', 'result'])
```

scanning/disa-alignment/shared.py
```diff
@@ -1,12 +1,47 @@
...
+            f"result='{self.result}', "
+            f"comparison_with_stig_ids={self.stig_ids_results}, "
+            f"final_result={self.final_result})"
+        )
+
+
+class ComparisonResult:
+    SAME = "SAME"
+    DIFFERENT = "DIFFERENT"
+    MISSING = "MISSING"
```
comps, line 43: Unless you print the `"SAME"`, `"DIFFERENT"`, etc. strings somewhere, using `enum` is cleaner and faster.

Do
```python
import enum
```
at the top, then

```suggestion
class ComparisonResult(enum.Enum):
    SAME = enum.auto()
    DIFFERENT = enum.auto()
    MISSING = enum.auto()
```
See https://docs.python.org/3/library/enum.html .

scanning/disa-alignment/shared.py
```diff
@@ -42,27 +77,121 @@ def disa_scan(host, ds, html, arf):
...
+        if same >= 1:
+            ssg_rule_result.final_result = ComparisonResult.SAME
+        elif different >= 1:
+            ssg_rule_result.final_result = ComparisonResult.DIFFERENT
+        elif missing >= 1:
+            ssg_rule_result.final_result = ComparisonResult.MISSING
+
+
+def get_disa_result_to_str(stig_ids_results):
+    if len(stig_ids_results) == 0:
```
comps, line 155: When checking if `dict` / `list` / etc. is empty, you can just use the boolean value of that object (internally mapped to `__bool__()`, you don't need to compare length to `0`.
```suggestion
    if stig_ids_results:
```

scanning/disa-alignment/shared.py
```diff
@@ -42,27 +77,121 @@ def disa_scan(host, ds, html, arf):
...
+        return "Empty"
+    ll = [f"{k}:{v}" for k, v in stig_ids_results.items()]
```
comps, line 157: To save memory, you don't need to convert everything to a list, only to "destroy" it in the `join()`, you can use generators to let `join()` consume one item at a time.
```suggestion
    ll = (f"{k}:{v}" for k, v in stig_ids_results.items())
```
More on https://wiki.python.org/moin/Generators .

PR #347, opened 2025-02-24
conf/find_invalid_waivers.py: new script to indetify invalid waivers

matusmarhefka:
Example output using results from the latest CaC/content stabilization (v0.1.76):
```
$ ./find_invalid_waivers.py test_results/rhel10/results.txt.gz test_results/rhel9/results.txt.gz test_results/rhel8/results.txt.gz
2025-02-24 17:26:57 find_invalid_waivers.py:161: lib.waive.collect_waivers:149: using /home/matus/tests/contest/conf/waivers for waiving
===============================================================
The following waivers are no longer valid, they either did not
match any test results or only matched the 'pass' test results:
===============================================================

/hardening/host-os/ansible/anssi_.+/timer_logrotate_enabled
/hardening/host-os/ansible/anssi_.+/timer_dnf-automatic_enabled
    True

/per-rule/.+/directory_permissions_var_log_audit/incorrect_value_0700.fail
/per-rule/.+/tftpd_uses_secure_mode/correct.pass
/per-rule/.+/dconf_gnome_lock_screen_on_smartcard_removal/wrong_value.fail
/per-rule/.+/tftpd_uses_secure_mode/wrong.fail
/per-rule/.+/file_ownership_var_log_audit_stig/correct_value_default_file.pass
/per-rule/.+/directory_permissions_var_log_audit/correct_value_0700.pass
    rhel == 9

/per-rule/.+/package_talk_removed/package-installed.fail
    rhel == 8

/per-rule/.+/audit_rules_unsuccessful_file_modification_openat_o_creat/o_creat_rules.pass
/per-rule/.+/audit_rules_unsuccessful_file_modification_openat_o_trunc_write/o_trunc.pass
/per-rule/.+/audit_rules_unsuccessful_file_modification_open_rule_order/ordered_filter.pass
/per-rule/.+/audit_rules_unsuccessful_file_modification_openat_rule_order/ordered_arch.pass
/per-rule/.+/audit_rules_unsuccessful_file_modification_open_rule_order/ordered_arch.pass
/per-rule/.+/audit_rules_unsuccessful_file_modification_openat_o_creat/o_creat_last.pass
/per-rule/.+/audit_rules_unsuccessful_file_modification_openat_rule_order/ordered_filter.pass
    rhel == 10

/scanning/disa-alignment/.*/harden_sshd_ciphers_opensshserver_conf_crypto_policy
    rhel == 9

/scanning/disa-alignment/[^/]+/auditd_audispd_configure_sufficiently_large_partition
    True

/hardening/host-os/oscap/[^/]+/service_nftables_disabled
    True

/hardening/anaconda/with-gui/cis_workstation_l[12]
    status == 'error'

/hardening/image-builder/stig
    rhel == 9 and status == 'error'

/scanning/disa-alignment/.*/sysctl_kernel_core_pattern
    rhel == 8

/hardening/host-os/oscap/[^/]+/package_dnf-automatic_installed
/hardening/host-os/oscap/[^/]+/package_rsyslog-gnutls_installed
/hardening/host-os/oscap/[^/]+/timer_dnf-automatic_enabled
    True

/hardening/host-os/.*/(ospp|cui)/timer_dnf-automatic_enabled
    rhel == 8 or rhel == 9

/scanning/oscap-eval/ERROR
    rhel == 8 and note == 'E: oscap: Failed to convert OVAL state to SEXP, id: oval:ssg-state_file_groupowner_var_log_syslog_gid_4_0:ste:1.'

/static-checks/html-links/.+
    "failed: certificate has expired" in note

```

### DIFF
--- a/conf/find_invalid_waivers.py
+++ b/conf/find_invalid_waivers.py
@@ -1,0 +1,161 @@
+#!/usr/bin/python3
+"""
+This is a standalone script which processes provided results.txt.gz files
+and identifies invalid waivers. The waiver is invalid if it:
+ - did not match any test results,
+ - or only matched the 'pass' test results.
+
+The identified invalid waivers are printed to the standard output.
+"""
+
+import re
+import sys
+import gzip
+import pathlib
+import argparse
+import textwrap
+
+# add the parent directory to the sys.path so we can import from the lib directory
+sys.path.append(str(pathlib.Path(__file__).resolve().parent.parent))
+from lib import waive, versions
+
+_sections_cache = None
+
+
+def find_regexes_in_list(regexes_matched_list, regexes):
+    for index, (list_regexes, _) in enumerate(regexes_matched_list):
+        if list_regexes == regexes:
+            return index
+    return None
+
+
+def add_regexes_to_list(regexes_matched_list, regexes, non_pass_result):
+    index = find_regexes_in_list(regexes_matched_list, regexes)
+    if index is not None:
+        # tuple with the regexes is already in the list, the boolean value
+        # needs to be modified only in case we matched a non-pass test result
+        # and it is not already marked as such in the existing tuple
+        if non_pass_result is True and regexes_matched_list[index][1] is False:
+            regexes_matched_list[index] = (regexes_matched_list[index][0], non_pass_result)
+    else:
+        # tuple with the regexes not found, add it to the list
+        regexes_matched_list.append((regexes, non_pass_result))
+
+
+def unwaive_note(waive_text, note):
+    return note.lstrip(waive_text).rstrip(')')
+
+
+def match_result_mark_waiver(regexes_matched_list, version, arch, status, name, note):
+    """This function is an updated version of the match_result() function from the lib/waive.py."""
+    version = versions._Rhel(version=version, id='rhel')
+
+    # make sure "'something' in name" always works
+    if name is None:
+        name = ''
+    if note is None or note == '[]':
+        note = ''
+
+    if '(waived pass)' in note:
+        note = unwaive_note('(waived pass)', note)
+    elif '(waived fail)' in note:
+        note = unwaive_note('(waived fail)', note)
+        status = 'fail'
+    elif '(waived error)' in note:
+        note = unwaive_note('(waived error)', note)
+        status = 'error'
+
+    objs = {
+        'status': status,
+        'name': name,
+        'note': note,
+        'arch': arch,
+        'rhel': version,
+        'oscap': '',
+        'env': '',
+        'Match': waive.Match,
+    }
+
+    for section in _sections_cache:
+        if any(x.fullmatch(name) for x in section.regexes):
+            ret = eval(section.python_code, objs, None)
+
+            if not isinstance(ret, waive.Match):
+                if not isinstance(ret, bool):
+                    raise RuntimeError(f"waiver python code did not return bool or Match: {ret}")
+                ret = waive.Match(ret)
+
+            if ret:  # both regex and python code matched
+                add_regexes_to_list(regexes_matched_list, section.regexes, bool(status != 'pass'))
+
+
+def load_and_process_results(file, regexes_matched_list):
+    with gzip.open(file, 'rt') as f:
+        for line in f:
+            line = line.strip()
+            version, arch, status, name, note = line.split('\t')
+
+            if status not in ['pass', 'fail', 'error', 'warn']:
+                continue
+
+            match_result_mark_waiver(regexes_matched_list, version, arch, status, name, note)
+
+
+def main(result_file_list):
+    regexes_matched_list = []
+    """
+    [
+        # list of tuples containing regexes matching some test result and a Boolean value
+        # set to True if regexes matched any non-pass test result, otherwise False
+        (set(regexes), True/False),
+        (set(regexes), True/False),
+        ...
+    ]
+    """
+    global _sections_cache
+    _sections_cache = list(waive.collect_waivers())
+
+    for file in result_file_list:
+        load_and_process_results(file, regexes_matched_list)
+
+    print("===============================================================\n"
+          "The following waivers are no longer valid, they either did not\n"
+          "match any test results or only matched the 'pass' test results:\n"
+          "===============================================================\n")
+    for section in _sections_cache:
+        index = find_regexes_in_list(regexes_matched_list, section.regexes)
+        # if the section did not match any test results or only matched the 'pass' test results:
+        if index is None or regexes_matched_list[index][1] is False:
+            python_source = next(
+                (i.python_source for i in _sections_cache if i.regexes == section.regexes),
+                None,
+            )
+            # if "is_centos()" is in python_src_code variable and there is no "or"
+            # logical operator the section is only applicable for centos so skip it
+            or_operator_in_py_code = bool(re.search(r'\s+or\s+', python_source) is not None)
+            centos_section = bool('is_centos()' in python_source and not or_operator_in_py_code)
+            if not centos_section:
+                for regex in section.regexes:
+                    print(regex.pattern)
+                print(f"    {python_source}\n")
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(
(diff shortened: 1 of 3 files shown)


conf/find_invalid_waivers.py
```diff
@@ -0,0 +1,161 @@
...
+import gzip
+import pathlib
+import argparse
+import textwrap
+
+# add the parent directory to the sys.path so we can import from the lib directory
+sys.path.append(str(pathlib.Path(__file__).resolve().parent.parent))
+from lib import waive, versions
+
+_sections_cache = None
```
comps, line 22: I don't think this global variable is needed.

It exists in `lib.waive` because we don't want passing it around in test code, we want the test to be able to just call
```python
results.report(...)
```
without passing some parsed list of waivers around its logic.

But that problem doesn't exist here in your code, so you can IMHO leave it as a normal variable inside `main()`.

lib/versions.py
```diff
@@ -60,10 +60,14 @@ def __str__(self):
...
-        _update_os_release()
+    def __init__(self, version=None, id=None):
         self._release_separator = '.'
-        v = _os_release['VERSION_ID'].split(self._release_separator)
+        if version and id:
+            v = version.split(self._release_separator)
+            _os_release['ID'] = id
+        else:
+            _update_os_release()
+            v = _os_release['VERSION_ID'].split(self._release_separator)
```
comps, line 70: I'm not super comfortable with the code (run in any possible instance of `class _Rhel`) modifying the global `_os_release` state.

I know it's used only from the waiver checking script, but I think a much cleaner solution would be to provide a "fake `rhel`" to the waiver python code, rather than trying to hack the one from `lib.versions`.

Ie.
```python
class FakeRhel(versions._RpmVerCmp):
    def __init__(self, fake_major, fake_minor):
        self.version = str(fake_major)
        self.release = str(fake_minor)
        self.major = fake_major
        self.minor = fake_minor

    def is_true_rhel(self):
        return True  # always assume we're on RHEL

    def is_centos(self):
        return False  # always assume we're on RHEL

    def __bool__(self):
        return True  # always assume we're on RHEL
```
and then you can simply instantiate it and pass that instance to waivers:
```python
   objs = {
        ...
        'rhel': FakeRhel(9, 6),
        ...
    }
```

comps (commented):
I think I would prefer this being in some new `tools` or `scripts` directory, not in `conf`, which is supposed to be importable by tests and `lib`.

conf/find_invalid_waivers.py
```diff
@@ -0,0 +1,161 @@
...
+    regexes_matched_list = []
+    """
+    [
+        # list of tuples containing regexes matching some test result and a Boolean value
+        # set to True if regexes matched any non-pass test result, otherwise False
+        (set(regexes), True/False),
+        (set(regexes), True/False),
+        ...
+    ]
+    """
```
comps, line 114: What is this docstring-like syntax?

I remember this being used for class attributes, but only unofficially - `pydoc` / Sphinx doesn't AFAIK support it, so I would recommend just using a `#` comment above the list, or defining a `namedtuple` for each line, and documenting the `namedtuple` via a `#` comment if needed.

conf/find_invalid_waivers.py
```diff
@@ -0,0 +1,161 @@
...
+            line = line.strip()
+            version, arch, status, name, note = line.split('\t')
+
+            if status not in ['pass', 'fail', 'error', 'warn']:
+                continue
+
+            match_result_mark_waiver(regexes_matched_list, version, arch, status, name, note)
+
+
+def main(result_file_list):
```
comps, line 104: Why define a separate `main()` instead of having these contents directly under
```python
if __name__ == '__main__':
```
?

Not strictly against this, just curious.

conf/find_invalid_waivers.py
```diff
@@ -0,0 +1,161 @@
...
+    global _sections_cache
+    _sections_cache = list(waive.collect_waivers())
+
+    for file in result_file_list:
+        load_and_process_results(file, regexes_matched_list)
+
+    print("===============================================================\n"
+          "The following waivers are no longer valid, they either did not\n"
+          "match any test results or only matched the 'pass' test results:\n"
+          "===============================================================\n")
```
comps, line 124: This is something we hit with Jan as well.

The problem with split strings (`"..."` separated by spaces or newlines, NOT `,`-separated arguments) is that flake8/Ruff treats them as a **single argument**, which they technically are.

This means that
```python
print(
    "foo "
    "bar"
)
```
triggers an error, because it sees a multi-line function call, but doesn't see `,` on each line, which cannot be there, because it's not multiple `print` arguments, it's one split argument.

The counter-intuitive, but probably most correct, solution is to add one `,` one the last line of the split string, satisfying the condition that a multi-like function call must have `,`-terminated arguments.

```suggestion
    print(
        "===============================================================\n"
        "The following waivers are no longer valid, they either did not\n"
        "match any test results or only matched the 'pass' test results:\n"
        "===============================================================\n",
    )
```
This makes it consistent with other multi-line function calls.

conf/find_invalid_waivers.py
matusmarhefka: Yes, but I can either merge it or rename the `main()` function to something like `get_invalid_waivers`.

conf/find_invalid_waivers.py
```diff
@@ -0,0 +1,161 @@
...
+                (i.python_source for i in _sections_cache if i.regexes == section.regexes),
+                None,
+            )
+            # if "is_centos()" is in python_src_code variable and there is no "or"
+            # logical operator the section is only applicable for centos so skip it
+            or_operator_in_py_code = bool(re.search(r'\s+or\s+', python_source) is not None)
+            centos_section = bool('is_centos()' in python_source and not or_operator_in_py_code)
+            if not centos_section:
+                for regex in section.regexes:
+                    print(regex.pattern)
```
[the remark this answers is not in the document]
matusmarhefka: Using `FakeCentos` approach would be more difficult to implement as python code under regexes can be more complex. I consider the current filtering good enough, it will remove most of the CentOS regexes, it will only keep things like:
```
regex
    rhel.is_centos() and note == 'foo or bar'
```
but that is fine imo, reviewer can skip such corner cases.

scripts/find_invalid_waivers.py
```diff
@@ -0,0 +1,185 @@
...
+        # and it is not already marked as such in the existing tuple
+        if non_pass_result is True and regexes_matched_list[index].non_pass_result is False:
+            regexes_matched_list[index] = MatchedWaivers(regexes_matched_list[index].regexes, True)
+    else:
+        # tuple with the regexes not found, add it to the list
+        regexes_matched_list.append(MatchedWaivers(regexes, non_pass_result))
+
+
+def unwaive_note(waive_text, note):
+    return note.lstrip(waive_text).rstrip(')')
```
comps, line 75: Huh? Why are you removing all characters from ie. `(waived pass)`?
Note that `.lstrip()` is not `.removeprefix()` (python >= 3.9) or `re.sub()` (for older pythons).

Maybe this should be
```python
# the extra space is present when the note contains more text
return note.removeprefix(waive_text).removeprefix(' ')
```
?

scripts/find_invalid_waivers.py
```diff
@@ -0,0 +1,185 @@
...
+    print(
+        "===============================================================\n"
+        "The following waivers are no longer valid, they either did not\n"
+        "match any test results or only matched the 'pass' test results:\n"
+        "===============================================================\n",
+    )
+    for section in _sections_cache:
+        index = find_regexes_in_list(regexes_matched_list, section.regexes)
+        # if the section did not match any test results or only matched the 'pass' test results:
+        if index is None or regexes_matched_list[index].non_pass_result is False:
```
comps, line 154: You spend all the effort of recording separately whether the waiver matched + whether it didn't match on pass, and then you seem to use only one boolean state here.

Wouldn't it make more sense to throw away the True/False (whether a section matched a pass) and only record sections that didn't match pass in the first place?

IOW putting pass-matching ones into the same bucket as never-matching ones, without making a distinction?

Issue #256, opened 2024-08-18
Break up `LastTest.log` of `/static-checks/rpmbuild-ctest`


comps:
The log can be several MBs (typically 6-8MB) and contains results for all tests executed, even passed ones.

Modify the test to provide a sectioned-off log file for each of the sub-results, like `/per-rule` does, so that logs for passed tests can be discarded if needed.

Ie. instead of `ctest_results`, iterate directly over `LastTest.log`, matching on sections like
```
18/43 Testing: test-generate_contributors
18/43 Test: test-generate_contributors
Command: "/usr/bin/env" "PYTHONPATH=/root/content:../.." "/usr/bin/python3" "/root/content/utils/generate_contributors.py" "--dry-run"
Directory: /root/content/build/tests
"test-generate_contributors" start time: Aug 17 23:18 CEST
Output:
----------------------------------------------------------
```
as a starting barrier, and
```
<end of output>
Test time =   0.02 sec
----------------------------------------------------------
Test Passed.
"test-generate_contributors" end time: Aug 17 23:18 CEST
"test-generate_contributors" time elapsed: 00:00:00
----------------------------------------------------------
```
as the ending barrier, extracting test name, pass/fail/etc. from it, and submitting any test output between the barriers as a log file to `results.report()`.

comps:
Also solve: https://github.com/RHSecurityCompliance/contest/issues/93

PR #355, opened 2025-03-10
print last 100 lines of scapval.sh on error

comps:
(Also update to full 1.3.6 release of scapval.)

### DIFF
--- a/static-checks/nist-validation/test.py
+++ b/static-checks/nist-validation/test.py
@@ -3,14 +3,13 @@
 import io
 import zipfile
 import requests
-import subprocess
 import xml.etree.ElementTree as ET
 from pathlib import Path
 
 from lib import util, results
 
 
-url = 'https://github.com/RHSecurityCompliance/contest-data/raw/main/data/scapval-1.3.6-rc3.zip'
+url = 'https://github.com/RHSecurityCompliance/contest-data/raw/main/data/scapval-1.3.6.zip'
 r = requests.get(url)
 r.raise_for_status()
 zip = zipfile.ZipFile(io.BytesIO(r.content))
@@ -30,7 +29,12 @@ for datastream in util.iter_datastreams():
         '-valresultfile', result_file,
         '-file', datastream,
     ]
-    util.subprocess_run(cmd, stdout=subprocess.DEVNULL, check=True)
+    with util.LastLinesBuffer(100) as f:
+        proc = util.subprocess_run(cmd, stdout=f, stderr=f)
+    if proc.returncode != 0:
+        util.log(f"scapval.sh output:\n----------\n{f.output.rstrip()}\n----------")
+        proc.check_returncode()
+
     tree = ET.parse(result_file)
     root = tree.getroot()
     for elem in root.findall('./nist:results/nist:base-requirement', ns):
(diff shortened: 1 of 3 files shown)


static-checks/nist-validation/test.py
```diff
@@ -30,7 +29,12 @@
         '-valresultfile', result_file,
         '-file', datastream,
     ]
-    util.subprocess_run(cmd, stdout=subprocess.DEVNULL, check=True)
+    with util.LastLinesBuffer(100) as f:
+        proc = util.subprocess_run(cmd, stdout=f, stderr=f)
```
matusmarhefka, line 33: Instead of implementing `LastLinesBuffer` wouldn't it be better to just do something like this:
```
util.subprocess_run(cmd, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, text=True)
if result.returncode != 0:
    lines = result.stdout.splitlines()
    print("\n".join(lines[-100:]))
    proc.check_returncode()
```

comps: No, because that would still read the entire stdout (it would be probably fine here with just 1-2 MB of output, but for a tool with GBs of output, it would eat GBs of memory). `LastLinesBuffer` discards any lines except for the last N ones due to the `collections.deque(maxlen=...)` limit, which works like a circular buffer.

There is unfortunately no functionality in `subprocess` that wouldn't read the full process output to memory. Both `.communicate()` and `PIPE` do that. :disappointed:

I figured the `LastLinesBuffer` might come in handy elsewhere too, but if you don't want to include because of this, that's fine. The `scapval.sh` output could honestly be printed out in its entirety, it isn't bigger than ie. `ansible-playbook` outputs we have elsewhere, I just thought it'd be an interesting use case.

matusmarhefka: Where would be the other use cases for `LastLinesBuffer` in contest? If we have any good candidates where it can be used let's keep it, otherwise I would discard it because I don't think it is necessary for this test.

comps: `ansible-playbook` output could be a good candidate, but we often need to search the full output to see if ie. some rule got remediated at all, so it's probably best to keep it as a full log.

However Anaconda installations rarely / never have anything useful outside of the last ~50 lines, and they can fill up many MBs of output, so we could use `LastLinesBuffer` for them - we would lose the realtime VM installation output, but we use VMs for a lot of use cases, so it could cut down a lot on log sizes. I can push a second commit to this PR.

Also, https://github.com/RHSecurityCompliance/contest/issues/256 could be a potential use case, depending on how the log is structured. All I know is that it is big. There are many others (Image Builder composer output, some `podman` outputs, etc.) that could benefit from it, but the benefit probably wouldn't be huge.

...
OTOH, we could just drop `LastLinesBuffer` and have full output of `scapval.sh` without significant issues, `ansible-playbook` could be gzipped, and VM install outputs are already smaller than with verbose systemd (that we had because of RHEL-7).

I guess it's up to you which would you prefer.

matusmarhefka: Okay, let's keep it then. We can consider if we would like to use it also with VM install outputs later.

matusmarhefka (approved):

PR #358, opened 2025-03-10
Add waiver for package_bind_removed with ansible on RHEL 9 per rule test

ggbecker:


conf/waivers/productization
```diff
@@ -131,7 +131,7 @@
     True
 
 # https://github.com/ComplianceAsCode/content/issues/13107
-/per-rule/.+/oscap/package_bind_removed/package-installed.fail
+/per-rule/.+/(oscap|ansible)/package_bind_removed/package-installed.fail
```
comps, line 134: There are only oscap and ansible tests, so I would shorten it (like waivers above it) to
```suggestion
/per-rule/.+/package_bind_removed/package-installed.fail
```

PR #362, opened 2025-03-11
Add waiver for rootfiles_configured in RHEL9

ggbecker:
https://github.com/ComplianceAsCode/content/issues/13160

### DIFF
--- a/conf/waivers/productization
+++ b/conf/waivers/productization
@@ -89,6 +89,9 @@
 # https://github.com/ComplianceAsCode/content/issues/13021
 /scanning/disa-alignment/.*/directory_permissions_sshd_config_d
     rhel == 9
+# https://github.com/ComplianceAsCode/content/issues/13160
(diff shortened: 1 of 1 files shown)


conf/waivers/productization
```diff
@@ -22,6 +22,10 @@
 /per-rule/.+/service_telnet_disabled/service_disabled.pass
     rhel == 9
 
+# https://github.com/ComplianceAsCode/content/issues/13160
+/scanning/disa-alignment/.+/rootfiles_configured
+    rhel == 9
+
```
comps, line 28: A small nitpick - we try to keep all the disa-alignments in one block, to separate them from other rules. Consider also appending this to one of the other `rhel == 9` sections in `disa-alignment`, though creating a new one is fine too.
```suggestion
```

ggbecker: Ok, I moved it to the end of the disa alignment block and removed the space. I couldn't find any existing waiver that could be merged with this new one though.

PR #366, opened 2025-03-13
OPENSCAP-5421 - find_invalid_waivers.py: recognize individual invalid waivers within a section

matusmarhefka:
The script was able to identify only the whole sections of invalid waivers, therefore sections which contained at least one valid waiver were ignored. This commit improves the script so it is also possible to identify individual invalid waivers within sections which also contain valid waivers.

### DIFF
--- a/scripts/find_invalid_waivers.py
+++ b/scripts/find_invalid_waivers.py
@@ -14,10 +14,17 @@ import gzip
 import pathlib
 import argparse
 import textwrap
+import collections
 
 # add the parent directory to the sys.path so we can import from the lib directory
 sys.path.append(str(pathlib.Path(__file__).resolve().parent.parent))
 from lib import waive, versions
+
+
+MatchedWaiver = collections.namedtuple(
+    'MatchedWaiver',
+    ['regex', 'python_source'],
+)
 
 
 class _FakeRhel(versions._RpmVerCmp):
@@ -63,9 +70,7 @@ def match_result_mark_waiver(regexes_matched_list, version, arch, status, name, 
     if note is None or note == '[]':
         note = ''
 
-    if '(waived pass)' in note:
-        note = unwaive_note('(waived pass)', note)
-    elif '(waived fail)' in note:
+    if '(waived fail)' in note:
         note = unwaive_note('(waived fail)', note)
         status = 'fail'
     elif '(waived error)' in note:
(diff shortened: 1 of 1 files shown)


comps:
I don't have any comments to the code, LGTM.

I still think we should have handled CentOS by trying to match against `class _FakeCentos` rather than grepping the python source for patterns. :smile:

matusmarhefka:
> I don't have any comments to the code, LGTM.
> 
> I still think we should have handled CentOS by trying to match against `class _FakeCentos` rather than grepping the python source for patterns. 😄

Even if you have `class _FakeCentos` which would match you still need to check if waiver is applicable only on CentOS of if there is other condition(s) matching also on RHEL. Honestly, I don't see how would `class _FakeCentos` help here, but maybe I am missing something?

comps:
> Even if you have `class _FakeCentos` which would match you still need to check if waiver is applicable only on CentOS of if there is other condition(s) matching also on RHEL. Honestly, I don't see how would `class _FakeCentos` help here, but maybe I am missing something?

If would avoid the python code parsing and interpreting `or` or whatever other python code the user could have used, it would just programmatically try matching RHEL and if that doesn't match, try CentOS, and if that doesn't match, then waiver (regexp) is unused.

No dealing with any python code syntax via regular expressions; you just set up the `eval` environment to look like CentOS and run the python code to see if it matches.

It would future proof it against more complex python code using `is_centos()`, but mainly establish clear design barriers of "we don't have special cases - it's python code, it behaves like python code, we don't care what's in the code, we just run it".

---
But yes, it would take slightly more effort to do than a simple regex hack, which is why I don't want to be anal about it. I guess it's how my brain works - I hate special cases and exceptions, because they prevent my brain from doing `-O2` on the code logic. :smile:

matusmarhefka:
But you would still need to parse the python code. None of the results script processes would match CentOS and to fake the match, you would need to parse python source code to fill the matching object, for example in:
```
rhel.is_centos() and arch = s390x and note == 'missing https://public.cyber.mil/stigs/downloads/?_dl_facet_stigs=operating-systems%2Cunix-linux'
```
you would need to obtain `arch` and `note` and then fill the data into globals dictionary for eval. That is why I decided not to do it, I don't think it is worth the hassle just because of 3 CentOS waivers we have + it would require more code than the 2 lines which currently filter out CentOS waivers.

PR #368, opened 2025-03-13
Add waiver to rootfiles_configured for RHEL8 as well.

ggbecker:


conf/waivers/productization
```diff
@@ -91,7 +91,7 @@
     rhel == 9
 # https://github.com/ComplianceAsCode/content/issues/13160
 /scanning/disa-alignment/.+/rootfiles_configured
-    rhel == 9
+    rhel == 8 or rhel == 9
```
comps, line 94: Is this something that would happen on RHEL-10 too, if we were checking for its alignment?

If so, then let's make it generic.
```suggestion
    True
```

ggbecker: The RHEL10 STIG profile selects this rule but it doesn't seem to be manifesting the issue. So it looks like the rhel8/9 here seems to be the right approach.

Issue #227, opened 2024-07-17
Get rid of `python3-dnf` and `python3-rpm` dependencies


comps:
These currently block us from using an arbitrary python version, as they are available only for the default RHEL python version.

Turns out getting rid of the dependencies might not be that hard:

* `lib/util/old_content.py` - `_available_ssg_versions()` uses it just as a one-shot comparison of RPM versions, something that can easily be done via `subprocess`:
  * as `rpm --eval "%{lua:print(rpm.vercmp('1.2-1', '2.0-1'))}"` natively, or
  * `rpmdev-vercmp 1.12 1.13` from `rpmdevtools`, or
  * `/usr/libexec/platform-python -c "..."` one-shot call of the RHEL-default python interpreter, [just like old Bash Contest](https://github.com/RHSecurityCompliance/contest/commit/adca23141b869b940f79d66c58153f0736901f1b#diff-6094db44698a706ea61acad15c76b4279e0690199ea5395650ea3fca30688d85R24-R34)
* `lib/versions.py` - this is really just a left-over from us wanting to compare SSG RPM versions in waivers; the logic is currently used only for RHEL version comparisons that could easily be turned to simple `int()` comparisons, incl. floating points
  * technically, there's the ability to compare `openscap-scanner` RPM version, but we never do that - we just use `versions.rhel` to work around a specific openscap version
* `lib/dnf.py` - this is actually a valid complex use case of querying something not available outside python API, however we could easily have a separate tool in `lib`, like we do for `pseudotty`, that would run under the default python interpreter and have some trivial CLI interface to extract the needed repo URLs from the DNF API

PR #375, opened 2025-03-24
OPENSCAP-5344 - Use the latest available python3 interpreter on given RHEL version

matusmarhefka:
Remove dependency on `python3-dnf` and `python3-rpm` to make it possible
to use the latest python3 version available in a given RHEL version.

Currently, we run tests on the following RHEL versions:
8.8, 8.10, 9.0, 9.2, 9.4, 9.5, 9.6 and 10.0.

RHEL and python3 versions (https://access.redhat.com/solutions/6879401):
- RHEL-10 defaults to python 3.12, there is only python3 RPM package
- RHEL-9 defaults to python 3.9, python3.11 RPM is available since RHEL-9.2
- RHEL-8 defaults to python 3.6, python3.11 RPM is available since RHEL-8.8

Effectively, we are limited by RHEL-9.0 where only python 3.9 is available
so contest can currently only use python features up to version 3.9.

Resolves https://github.com/RHSecurityCompliance/contest/issues/227

### DIFF
--- a/lib/dnf.py
+++ b/lib/dnf.py
@@ -3,7 +3,7 @@ import collections
 import tempfile
 import subprocess
 import requests
-import dnf
+import json
 from pathlib import Path
 
 from lib import util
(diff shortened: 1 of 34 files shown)


lib/dnf.py
```diff
@@ -12,41 +12,29 @@
...
-            continue
-        repo.load()
-        baseurl = repo.remote_location('/')
+    cmd = util.libdir / 'dnf_get_repos'
+    ret = util.subprocess_run(cmd, check=True, stdout=subprocess.PIPE, universal_newlines=True)
+    try:
+        json_data = json.loads(ret.stdout)
+    except json.JSONDecodeError as e:
+        util.log(f"Failed to decode JSON data from {cmd}: {e}")
+        return
```
comps, line 22: Probably `str(e)`, although it doesn't matter much.

But we definitely don't want to just log & return, we want to re-raise it to stop the test.

```suggestion
        raise RuntimeError(f"Failed to decode JSON data from {cmd}: {e}") from None
```

lib/dnf.py
```diff
@@ -12,41 +12,29 @@
...
+
+    for i in json_data:
```
comps, line 24: Let's not use single-letter variables on this big scope.

Maybe?
```suggestion
    for repo in json_data:
```

lib/dnf_get_repos
```diff
@@ -0,0 +1,53 @@
...
+        data = dict(repo.cfg.items(name))
+        repos_list.append(
+            {
+                'name': name,
+                'baseurl': baseurl,
+                'data': data,
+                'file': repo.repofile,
+            },
+        )
+    return repos_list
```
comps, line 49: It might be cleaner to just make `get_repos_dnf()` a generator, and give it to
```
print(json.dumps(get_repos_dnf()))
```
as `list(get_repos_dnf()`.

(it would remove the need to append to some list, just do `yield {'name': ... }`)

lib/dnf_get_repos
```diff
@@ -0,0 +1,53 @@
...
+
+
+if __name__ == '__main__':
+    print(json.dumps(get_repos_dnf()))
```
comps, line 53: `json` can write to stdout directly, no need for `print`, just use `dump` instead of `dumps`:
```
json.dump(list(get_repos_dnf()), sys.stdout)
```

lib/versions.py
```diff
@@ -17,60 +14,17 @@ def _update_os_release():
...
         else:
-            return self.version
-
-
-class _Rhel(_RpmVerCmp):
-    def __init__(self):
-        _update_os_release()
-        self._release_separator = '.'
-        v = _os_release['VERSION_ID'].split(self._release_separator)
+            v = version.split('.')
```
comps, line 23: I think `major, _, minor = version.partition('.')` would be safer here, because it would make `minor` into `None` on platforms where the version doesn't contain comma (compatibility of `lib` for Fedora would be nice, but not mandatory).

lib/versions.py
```diff
@@ -86,23 +40,67 @@ def is_centos():
     def __bool__():
         return _os_release['ID'] in ['rhel', 'centos']
 
+    @staticmethod
+    def _parse_version(version):
+        if isinstance(version, _Rhel):
+            return version
+        if not isinstance(version, str):
+            version = str(version)
```
comps, line 48: This is probably fine, but consider using integers everywhere and just do `str -> int` conversion when reading from `os-release` (or interfacing with `rpm` via `subprocess`), just to avoid the extra str/int confusion in classes.

main.fmf
```diff
@@ -8,13 +8,13 @@ require:
...
   - python3-requests
-  - python36-requests
+  - python3.11-requests
   - python3-pyyaml
-  - python36-pyyaml
-  - python3-rpm
-  - python36-rpm
+  - python3.11-pyyaml
+  - python3.11-urllib3
+  - python3-dnf
```
comps, line 17: Consider moving these under `require:` (or `require+:`) below, to the `adjust` secions.

comps:
When rebasing, please grep again for `test:` in fmf, https://github.com/RHSecurityCompliance/contest/pull/377 added a new test without `CONTEST_PYTHON`.

main.fmf
matusmarhefka: I think this won't be possible through `adjust` because `rhel-9` matches also all `rhel-9.Y` versions and, on the other hand, something like `distro == rhel-9 and distro ~!= rhel-9.0` won't match `distro=rhel-9` context (needed because of runcontest). So I just used `recommend+` instead of `require+` and the way it is now should properly set `CONTEST_PYTHON` variable based on distro.

comps: Sure, either way works. Thanks for trying!

comps:
This is LGTM, assuming it passes testing.

PR #377, opened 2025-03-24
Add scanning/oscap-debug/sysctl-oval-eval-parallel test

evgenyz:


scanning/oscap-debug/sysctl-oval-eval-parallel.py
```diff
@@ -0,0 +1,98 @@
...
+import subprocess
+import concurrent.futures
+
+from lib import util, results
+
+
+def run(fn):
+    handle = util.subprocess_Popen(['oscap', 'oval', 'eval', fn])
+    try:
+        rc = handle.wait(OSCAP_TIMEOUT)
```
comps, line 13: Maybe `util.subprocess_run(..., timeout=OSCAP_TIMEOUT)` would be more elegant?

scanning/oscap-debug/sysctl-oval-eval-parallel.py
```diff
@@ -0,0 +1,98 @@
...
+        return handle.pid, rc
+    except subprocess.TimeoutExpired:
+        return handle.pid, -1
+
+
+start_time = time.monotonic()
+
+# sysctl ovals only take about 1 second
+OSCAP_TIMEOUT = 5
```
comps, line 22: Probably move the constant to the top, above `def run()`, so the reader notices its definition before use in the function.

scanning/oscap-debug/sysctl-oval-eval-parallel.py
```diff
@@ -0,0 +1,98 @@
...
+        set logging enabled on
+        thread apply all bt
+        set logging enabled off
+    '''))
+
+with util.get_source_content() as content_dir:
+    util.build_content(
+        content_dir,
+        force=True,
+    )
```
comps, line 49: Why the explicit `force=True`?

Without it (and without extra build options), `build_content()` may save some time by reusing already-built content that's a superset of the (optional) options given to `build_content()`. In this case, I assume you don't care as any content building would have generated the OVAL files, so perhaps:
```suggestion
    util.build_content(content_dir)
```
That is, unless the OVAL files encode absolute paths inserted during content building, in which case `force=True` might be needed.

scanning/oscap-debug/sysctl-oval-eval-parallel.py
```diff
@@ -0,0 +1,98 @@
...
+                )
+
+                if gdb.returncode != 0:
+                    results.report(
+                        'warn',
+                        f'attempt:{attempt}',
+                        f"gdb returned {gdb.returncode}",
+                    )
+                    # something went wrong with gdb, let's try again
+                    continue
```
comps, line 78: I'm pretty sure this code won't ever run, because `check=True` above raises an exception if `gdb` somehow fails.

(Which is probably the better solution if `gdb` somehow failing is a "never happens" occurrence.)

scanning/oscap-debug/sysctl-oval-eval-parallel.py
evgenyz: I need pid from process handle.

scanning/oscap-debug/sysctl-oval-eval-parallel.py
evgenyz: Yeah.

scanning/oscap-debug/sysctl-oval-eval-parallel.py
evgenyz: I was just copy-pasting from other test. Most likely I don't need it, any content should do the job.

scanning/oscap-debug/sysctl-oval-eval-parallel.py
evgenyz: This is for the off-chance of something other than hang-up happening, and gdb not being able to capture the trace and core. I guess I'll just remove the `check=True` part.

evgenyz:
BTW, I executed the test on RHEL9 machine. It was able to perform 205575 invocations. No hang-ups.

Issue #93, opened 2024-01-25
`/static-checks/rpmbuild-ctest` log eats a lot of space


comps:
The `LastTest.log` from `/static-checks/rpmbuild-ctest` has ~ 20MB of text, mainly due to the first 3 tests:
```
< 3 lines here >
1/53 Testing: sanity-ssg-tables-table-rhel9-cces.html
< 17000 lines here >
2/53 Testing: sanity-ssg-tables-table-rhel9-srgmap.html
< 23000 lines here >
3/53 Testing: sanity-ssg-tables-table-rhel9-srgmap-flat.html
< 63000 lines here >
4/53 Testing: unique-cces
< 1100 lines here >
```
We might want to either limit their output in CaC/content, or somehow trim them in the test.

PR #387, opened 2025-04-04
OPENSCAP-5346 - /static-checks/rpmbuild-ctest: break up LastTest.log

matusmarhefka:
This commit breaks up the `LastTest.log` into separate sections per each CTest test. Log sections of the passing tests are discarded, log sections of the failing tests are uploaded as test artifacts.

Resolves https://github.com/RHSecurityCompliance/contest/issues/93 and https://github.com/RHSecurityCompliance/contest/issues/256


static-checks/rpmbuild-ctest/test.py
```diff
@@ -27,14 +27,42 @@
...
+        state = 'start'
+        test_output = []
+
         for line in f:
-            result_match = result_regex.search(line)
-            if result_match:
-                test_name = result_match.group(1)
-                result = 'pass' if 'Passed' in line else 'fail'
-                results.report(result, test_name)
+            line = line.rstrip('\n')
```
comps, line 35: It seems you do `test_output.append(line)` in every single state, maybe put it here instead?
```suggestion
            line = line.rstrip('\n')
            test_output.append(line)
```

static-checks/rpmbuild-ctest/test.py
```diff
@@ -27,14 +27,42 @@
...
+
+            if state == 'output':
+                test_output.append(line)
+                if '<end of output>' in line:
```
comps, line 39: If the entire line from `^` to `$` is `<end of output>`, then I'd just do `line == '<end of output>'` to avoid matching on the string somewhere in valid test output (probably unlikely, so this is really a nitpick).

static-checks/rpmbuild-ctest/test.py
```diff
@@ -27,14 +27,42 @@
...
+
+            if state == 'after_output' and re.match(r'^-{10,}\s*$', line):
+                test_output.append(line)
+                state = 'result'
+                continue
+
+            if (
+                state == 'start'
+                and (m := re.fullmatch(r'^[0-9]+/[0-9]+\s+Testing:\s+(.+)$', line))
+            ):
```
comps, line 51: This is valid and you can keep it like this, but consider splitting it into two conditions to make it (IMHO) look better:
```suggestion
            if state == 'start':
                if m := re.fullmatch(r'^[0-9]+/[0-9]+\s+Testing:\s+(.+)$', line):
```
This is not possible every time, but here you do `continue` anyway, which breaks up the flow, so it can be done here.

static-checks/rpmbuild-ctest/test.py
matusmarhefka: Only lines in the expected states are appended because otherwise we would get some overlaps between tests (there are some additional lines we don't care about).

PR #424, opened 2025-07-15
Add a test for Ansible check mode

jan-cerny:
This test will test whether our Ansible Playbooks can run in check mode (`ansible-playbook --check`).

This test serves as a support for PR https://github.com/ComplianceAsCode/content/pull/13703.

This test will test if that PR resolves https://issues.redhat.com/browse/OPENSCAP-5480

### DIFF
--- a/scanning/host-os/ansible-check/check-mode/main.fmf
+++ b/scanning/host-os/ansible-check/check-mode/main.fmf
@@ -13,8 +13,6 @@ require+:
 recommend+:
   # needed for the ini_file ansible plugin, and more
   - rhc-worker-playbook
-  # individual rule playbooks
-  - scap-security-guide-rule-playbooks
 adjust+:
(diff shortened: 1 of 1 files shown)


comps:
Not sure if you're looking for feedback yet (PR in Draft), but:
* This probably belongs in `/scanning` (pehaps `/scanning/host-os/ansible-check` or so?), not `/static-checks` (which are about content-specific small checks that don't typically depend on host OS environment)
    * Maybe I would even make it a "check only" variant of `/hardening/host-os/ansible`, re-using most of the test code, basically just adding a new `check-only.fmf` like we have for `uefi.fmf` or `with-gui.fmf`
* The RpmPack code is probably not needed for non-destructive checking

jan-cerny:
I have remove individual playbook dependency.

matusmarhefka (approved):

PR #448, opened 2025-08-18
Create test for day 2 hardening of bootc systems

jan-cerny:


jan-cerny:
I have resolved the review suggestions.

comps:
Just a nitpick, but why call it `day2-upgrade`? .. We already have `/hardening/oscap/old-new` which seems to use exactly the same logic AFAICT, so this could be called `/hardening/container/old-new` for consistency.  
@matusmarhefka ?

hardening/container/day2-upgrade/test.py
```diff
@@ -0,0 +1,154 @@
...
+    COPY remediation-new-ds.xml /root/
+    # install and run oscap-im to harden the image
+    RUN dnf -y install openscap-utils
+    RUN oscap-im --profile '{profile}' \
+        --results-arf /etc/remediation-new-arf.xml /root/remediation-new-ds.xml
+    RUN bootc container lint
+''')
+cfile2.add_ssh_pubkey(guest.ssh_pubkey)
+cfile2.write_to('Containerfile')
+podman.podman('image', 'build', '--tag', 'contest-hardened-new', '.')
```
comps, line 82: This seems like unnecessary duplication just for "old" vs "new".

Have a look at how this is done in the other `old-new` test: https://github.com/RHSecurityCompliance/contest/blob/f045e57af1066d276ce24a2da47e51143b893386/hardening/oscap/old-new/test.py#L29-L46

Perhaps this code could also be moved to a function, which would then take filename arguments that could be specified as
```python
    build_image('remediation-old.xml', 'remediation-arf-old.xml', 'remediation-arf-old2.xml')
    build_image('remediation-new.xml', 'remediation-arf-new.xml', 'remediation-arf-new2.xml')
```
or something like that?

jan-cerny:
I have extract function build_image, I have removed the comment and I have renamed test

jan-cerny:
In the second image t's necessary to put it elsewhere than to /root because when upgrading a running Image Mode system the /root is kept and isn't overwritten by the contents of the image. The reason is that in Image Mode, the /root is a symlink to some directory in the /var file system and /var file system is designed to be resistant from image upgrades.

comps:
I meant something like this:
```diff
diff --git a/hardening/container/old-new/test.py b/hardening/container/old-new/test.py
index f0f88e8..573196b 100755
--- a/hardening/container/old-new/test.py
+++ b/hardening/container/old-new/test.py
@@ -7,43 +7,6 @@ from pathlib import Path
 from lib import results, oscap, versions, virt, podman, util
 from conf import remediation
 
-
-def build_image(image_version):
-    # prepare a Container file for making a hardened image using the data stream
-    if image_version == 'old':
-        data_stream = 'remediation-old-ds.xml'
-        arf = 'remediation-old-arf.xml'
-        with util.get_old_datastream() as old_ds:
-            oscap.unselect_rules(old_ds, data_stream, remediation.excludes())
-    elif image_version == 'new':
-        data_stream = 'remediation-new-ds.xml'
-        arf = 'remediation-new-arf.xml'
-        oscap.unselect_rules(
-            util.get_datastream(), data_stream, remediation.excludes())
-    else:
-        raise ValueError("image_version must be 'old' or 'new'")
-    cfile = podman.Containerfile()
-    cfile += util.dedent(fr'''
-        FROM {src_image}
-        # install testing-specific RpmPack
-        COPY contest-pack.rpm /root/.
-        RUN dnf -y install /root/contest-pack.rpm && rm -f /root/contest-pack.rpm
-        # copy over testing-specific datastream
-        COPY {data_stream} /root/
-        # install and run oscap-im to harden the image
-        RUN dnf -y install openscap-utils
-        RUN oscap-im --profile '{profile}' \
-            --results-arf /{arf} /root/{data_stream}
-        RUN bootc container lint
-    ''')
-    cfile.add_ssh_pubkey(guest.ssh_pubkey)
-    cfile.write_to('Containerfile')
-
-    podman.podman('pull', src_image)
-    podman.podman(
-        'image', 'build', '--tag', 'contest-hardened-' + image_version, '.')
-
-
 # prepare the host
 virt.Host.setup()
 
@@ -74,8 +37,39 @@ pack.add_sshd_late_start()
 with pack.build() as pack_binrpm:
     shutil.copy(pack_binrpm, 'contest-pack.rpm')
 
-build_image('old')
-build_image('new')
+# prepare old/new datastreams used for remediation
+with util.get_old_datastream() as old_ds:
+    oscap.unselect_rules(old_ds, 'remediation-old-ds.xml', remediation.excludes())
+oscap.unselect_rules(util.get_datastream(), 'remediation-new-ds.xml', remediation.excludes())
+
+
+# prepare a Container file for making a hardened image using the data stream
+def build_image(data_stream, results_arf, dst_image):
+    cfile = podman.Containerfile()
+    cfile += util.dedent(fr'''
+        FROM {src_image}
+        # install testing-specific RpmPack
+        COPY contest-pack.rpm /root/.
+        RUN dnf -y install /root/contest-pack.rpm && rm -f /root/contest-pack.rpm
+        # copy over testing-specific datastream
+        COPY {data_stream} /root/
+        # install and run oscap-im to harden the image
+        RUN dnf -y install openscap-utils
+        RUN oscap-im --profile '{profile}' --results-arf '{results_arf}' '{data_stream}'
+        RUN bootc container lint
+    ''')
+    cfile.add_ssh_pubkey(guest.ssh_pubkey)
+    cfile.write_to('Containerfile')
+
+    podman.podman('pull', src_image)
+    podman.podman('image', 'build', '--tag', dst_image, '.')
+
+
+# take remediation datastreams from CWD, as generated above,
+# store them in / inside the image because ??? why when /root is not overwritten ???
+# and build contest-hardened-* tagged images
+build_image('remediation-old-ds.xml', '/remediation-old-arf.xml', 'contest-hardened-old')
+build_image('remediation-new-ds.xml', '/remediation-new-arf.xml', 'contest-hardened-new')
 
 # pre-create a directory (inside GUEST_IMG_DIR) for storing the
 # hardened image, built by bootc-image-builder
```

The goal isn't to move everything old/new into a separate function, only to then add more old/new conditions inside that function, my suggestion was to move everything **that is identical for old/new** (a.k.a. is version-agnostic) into a shared function, and just call that function, giving it inputs/outputs as static arguments.  
The static arguments are there (instead of magically substituting old/new inside the function) to ease debugging - it's a lot more obvious what the function is doing and what places it's touching.

The idea is that the function itself doesn't know if it's doing old or new - it's a pure algorithm, that difference remains in the parent context.

For the same reason, moving the `unselect_rules()` inside it doesn't make sense - the code is very different for old vs new, so putting it in a shared function loses any benefit.

In the diff, I also moved the function down, because it uses global variables and I think it's bad practice to define a function using them **before** defining those variables, even if it works thanks to python dynamic evaluation.

Finally, I still don't understand why it's bad to store ARF results in `/root`:

> In the second image it's necessary to put it elsewhere than to /root because when upgrading a running Image Mode system the /root is kept and isn't overwritten by the contents of the image. The reason is that in Image Mode, the /root is a symlink to some directory in the /var file system and /var file system is designed to be resistant from image upgrades.

This seems to me like a good thing - if `/root` is preserved and not erased on upgrade, we can store the results there and copy them out later, no? The presence of `/root/remediation-old-arf.xml` after upgrade doesn't hurt anything. Why would we copy anything to places which get overwritten/deleted by the upgrade?

matusmarhefka:
@comps The problem is that `/root` of the old image is preserved, so after you upgrade to the new image, you won't see the ARF results from the new image under `/root`, only the original old ARF file (because home dirs are preserved with upgrades). So @jan-cerny solved it by placing ARF results under `/` (which gets overwritten) and copying the old ARF results before doing the image upgrade through `bootc switch`.

comps:
@matusmarhefka Ah, so the original old `/root` is preserved and the "new" ARF is **not** appended to it, even if explicitly mentioned in the `Containerfile`? So the hack is to store "old" ARF in `/root` (so it is accessible in the final OS), but store the "new" ARF in `/`, which would get overwritten by a second `bootc switch`, but since we don't do that, we can just copy it out from `/`, right?

In that case the code looks good to me, assuming my temporary `??? why when /root is not overwritten ???` comment that Jan copy/pasted is adjusted to explain the `/` vs `/root` in a way similar to how you did (to an average test developer).

matusmarhefka:
Testing passed, merging.

PR #449, opened 2025-08-25
New test /scanning/audit-rules-syscalls-grouping

matusmarhefka:
https://issues.redhat.com/browse/OPENSCAP-4502

### DIFF
--- a/lib/oscap.py
+++ b/lib/oscap.py
@@ -263,12 +263,14 @@ def unselect_rules(orig_ds, new_ds, rules):
     specified rules.
     """
     prefix = 'xccdf_org.ssgproject.content_rule_'
-    rules = (x if x.startswith(prefix) else prefix + x for x in rules)
+    # prefix rules once and store in a set for O(1) membership checks
+    prefixed_rules = {
+        (x if x.startswith(prefix) else prefix + x)
+        for x in rules
+    }
 
-    exprs = set()
-    for rule in rules:
-        exprs.add(re.compile(fr'<.*Rule.*id="{rule}'))
-        exprs.add(re.compile(fr'<.*select.*idref="{rule}'))
+    rule_def_re = re.compile(r'<[^>]*Rule[^>]*\bid="([^\"]+)"')
+    select_re = re.compile(r'<[^>]*select[^>]*\bidref="([^\"]+)"')
 
     new_ds = Path(new_ds)
     # remove a possible existing/old file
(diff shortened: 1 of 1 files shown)


hardening/audit-rules-grouping/test.py
```diff
@@ -0,0 +1,95 @@
...
+
+    with open(orig_ds) as orig_ds_f:
+        with open(new_ds, 'w') as new_ds_f:
+            for line in orig_ds_f:
+                if rule_line.search(line):
+                    if audit_rules.search(line):
+                        line = line.replace('selected="false"', 'selected="true"')
+                    else:
+                        line = line.replace('selected="true"', 'selected="false"')
+                new_ds_f.write(line)
```
comps, line 48: Maybe consider extending / optimizing `oscap.unselect_rules()` instead. Theoretically, you could do something like
```python
rules = list(oscap.global_ds().rules.keys())
oscap.unselect_rules(old_ds, new_ds, rules)
```
with the current `lib/` code already, but `oscap.unselect_rules()` tries to compile a regexp for every single rule and maybe that could be optimized by searches for XML structures that look like rule definitions + a second `in` check for each of the rule names.

plans/daily.fmf
```diff
@@ -32,6 +32,7 @@ discover:
       # run /per-rule as oscap only - this almost halves the runtime (for now)
       - /per-rule/oscap/[0-9]+$
       # the rest is cheap to run
+      - /hardening/audit-rules-grouping
```
comps, line 35: Are you sure we want to put this under `/hardening` when it (AFAICT) doesn't harden or even scan all profiles, reporting each rule separately?

hardening/audit-rules-grouping/test.py
```diff
@@ -0,0 +1,95 @@
...
+        for line in f:
+            # check if all syscalls from the group are present in the line
+            if all(re.search(pattern, line) for pattern in syscall_patterns):
+                match_found = True
+                util.log(line.strip())
+
+    if not match_found:
+        util.log("No matching audit rules found!")
+
+    return match_found
```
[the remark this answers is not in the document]
matusmarhefka: This is intentional, I want to log all audit rules matching the syscall group.

hardening/audit-rules-grouping/test.py
matusmarhefka: After unselecting all rules we would need to select the audit rules again so there would be more changes needed and we agreed it is not worth implementing for this single use case.

scanning/audit-rules-syscalls-grouping/test.py
```diff
@@ -0,0 +1,95 @@
...
+    audit_rules = re.compile(
+        r'<.*Rule.*id="xccdf_org.ssgproject.content_rule_audit_rules.*'
+        r'(_unsuccessful_file_modification|_dac_modification|'
+        r'_file_deletion_events|_kernel_module_loading|'
+        r'_networkconfig_modification|_time_adjtimex|'
+        r'_time_clock_settime|_time_settimeofday|_time_stime)',
+    )
+
+    with open(orig_ds) as orig_ds_f:
+        with open(new_ds, 'w') as new_ds_f:
```
comps, line 41: This could be also written as:
```suggestion
    with open(orig_ds) as orig_ds_f, open(new_ds, 'w') as new_ds_f:
```

comps:
LGTM with the last suggestion, or even without it. Feel free to merge if testing succeeds.

PR #460, opened 2025-09-15
Keep VM snapshots local to a test

### LINKED ISSUE #376: Keep VM snapshots local to a test
Currently, a test can leave a VM snapshot of a **running** VM behind (disk+RAM snapshot).

This has the advantage of it being quickly reused by subsequent test re-runs or runs of other tests, but there are several issues:
* If somebody starts up the VM using `./contest-sshvm`, it desyncs (corrupts) both the disk and RAM snapshot, which a subsequent test rerun might not catch
* It requires re-reading (re-discovering) IP address of the VM in a new test with a fresh `class Guest` instance
* It might break SSH if a test tries to run (reuse the snapshot) after `NETWORK_EXPIRY` hours (7 days)

But mainly, keeping the snapshot around is probably not good design - it keeps the VM in a "fragile" unstable state, which is also a problem for `CONTEST_LEAVE_GUEST_RUNNING`.

A better solution would be to **always** keep the VM shut down and **without snapshots**, but installed-and-ready, between tests.

A test could the easily
1. boot the VM up (1min?)
2. wait for boot to settle (30sec?)
3. Create a persistent ssh connection using `ControlMaster`
4. create a disk+RAM snapshot
5. do testing by (perhaps repeatedly) restoring the snapshot, using the persistent ssh connection
6. destroy / remove the disk and RAM snapshots

This would also allow hand-tuning of the VM environment for debugging - the test would always re-use an installed VM (given the right tag, eg. `Guest()` argument), so a human can add some debug/monitoring before rerunning the test. Kind of like manually set-up `CONTEST_CONTENT` with hand-merged PRs, etc.

matusmarhefka:
Resolves #376

Note: I also tried to setup a persistent ssh connection using `ControlMaster`, but the connection is interrupted after each snapshot revert so we would need to re-establish it every time we revert a snapshot. Therefore, I kept the original ssh implementation without the `ControlMaster`.

### DIFF
--- a/lib/virt.py
+++ b/lib/virt.py
@@ -37,16 +37,20 @@ instance to g.install().
 
 Example using snapshots:
 
+    import atexit
     import subprocess
-    import virt
+
+    from lib import virt
 
     virt.Host.setup()
     g = virt.Guest('gui')
 
     # reuse if it already exists from previous tests, reinstall if not
-    if not g.can_be_snapshotted():
+    if not g.is_installed():
         g.install()
-        g.prepare_for_snapshot()
+
+    g.prepare_for_snapshot()
+    atexit.register(g.cleanup_snapshot)
 
     with g.snapshotted():
         state = g.ssh('ls', '/root', stdout=subprocess.PIPE)
@@ -59,7 +63,7 @@ Example using snapshots:
 
 Example using plain one-time-use guest:
 
-    import virt
+    from lib import virt
 
     virt.Host.setup()
     g = virt.Guest()
@@ -353,7 +357,7 @@ class Guest:
     When instantiated, represents a guest (VM).
 
     Set a 'tag' (string) to a unique name you would like to share across tests
-    that use snapshots - the .can_be_snapshotted() function will return True
+    that use snapshots - the .is_installed() function will return True
     when it finds an already installed guest using the same tag.
     Tag-less guests can be used only for snapshotting within the same test
     and should not be shared across tests.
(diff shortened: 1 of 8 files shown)


lib/virt.py
```diff
@@ -596,21 +603,33 @@ def is_installed(self):
...
+            self.disk_path, self.disk_format = get_state_base_image_disk(self.name)
+            if current_disk_path != self.disk_path:
+                util.log(f"found {self.name} without reverted snapshot")
+                util.log(
+                    f"restoring {self.name} from {self.state_file_path}, "
+                    f"using original base image disk: {self.disk_path}",
+                )
+                self.destroy()
+                set_state_image_disk(self.state_file_path, self.disk_path, self.disk_format)
+                virsh('restore', self.state_file_path, check=True)
```
comps, line 623: I'm not sure I understand this part - the whole point of #376 is to **not** keep (disk or RAM) snapshots between tests. So I would imagine this code being something like "if there is any disk/ram snapshot, delete it" (it was probably left by an interrupted test).

Why are we restoring some "original" snapshot?

lib/virt.py
```diff
@@ -596,21 +603,33 @@ def is_installed(self):
...
+
+        # do guest first boot, let it settle and finish firstboot tasks
         self.start()
-        ip = wait_for_ifaddr(self.name)
-        wait_for_ssh(ip)
+        self.ipaddr = wait_for_ifaddr(self.name)
+        wait_for_ssh(self.ipaddr)
```
comps, line 628: I'm not sure this is a good idea, we already do this in both `booted()` and `snapshotted()` later, and duplicating it here is IMHO somewhat confusing.

Maybe we could do it only here and not in `snapshotted()`, since we now ALWAYS need to `prepare_for_snapshot()` before using `snapshotted()`?

lib/virt.py
```diff
@@ -676,12 +700,16 @@ def snapshotted(self):
...
+            raise RuntimeError(
+                f"guest {self.name} not prepared for snapshotting, "
+                "prepare_for_snapshot() needs to be used first",
+            )
         self._restore_snapshotted()
-        if not self.ipaddr:
-            self.ipaddr = wait_for_ifaddr(self.name)
-            wait_for_ssh(self.ipaddr)
+        self.ipaddr = wait_for_ifaddr(self.name)
+        wait_for_ssh(self.ipaddr)
```
comps, line 712: I believe this was an important optimization for fast snapshots - with week+ lease time, it's unlikely the VM changes IP address during test runtime, and waiting for `virsh domifaddr` takes extra time on every restore.

comps (changes requested):

PR #462, opened 2025-09-15
Skip profiles not in datastream for `rule-identifiers`

comps:
This should fix missing OSPP in downstream.


matusmarhefka (approved):
LGTM, feel free to merge if testing passes.

PR #477, opened 2025-10-22
Move no CCN identifiers to permanent waivers

Mab879:
This is expected as CCN is only for RHEL


.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+{
+    "python-envs.defaultEnvManager": "ms-python.python:system",
+    "python-envs.pythonProjects": []
+}
```
comps: What is this code for? Is it intentional? Or should we add `.vscode` to `.gitignore`?

PR #486, opened 2025-11-05
Remove Beaker-native result reporting

### LINKED ISSUE #228: Get rid of Beaker-native result reporting
In ba9fcc7058d924d2c9b85ccb1bf5ecc25c9c08ea, we worked around `tcms-results`'s inability to treat `None` / `Skip` / etc. as a non-failure state.

In earlier commits, ie. 0816e6156a7797da5098345a7d7200a095c39f40 we at least added the original result to the result name (message), but it would be best if we could just use native Beaker statuses, like before.

With TCMS gone (also https://github.com/RHSecurityCompliance/contest/issues/151), we can restore the Beaker-native reporting API.

Alternatively, we may just remove it and rely on TMT only.

matusmarhefka:
Resolves #228

### DIFF
--- a/lib/results.py
+++ b/lib/results.py
@@ -1,21 +1,18 @@
 """
 Functions and helpers for result management when working with
  - TMT (https://github.com/teemtee/tmt)
- - Beaker (https://beaker-project.org/docs/server-api/index.html)
+ - ATEX (https://github.com/RHSecurityCompliance/atex)
 
 TMT has a 'result:custom' feature (in test's main.fmf), which allows us to
 supply completely custom results as a YAML file, and TMT will use it as-is
 to represent a result from the test itself, and any results "under" it,
 effectively allowing a test to report more than 1 result.
-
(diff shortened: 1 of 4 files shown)


lib/util/environment.py
```diff
@@ -51,11 +50,5 @@ def get_test_name():
     name = os.environ.get('TMT_TEST_NAME')
     if name:
         return name
-    # under Restraint (Beaker, OSCI, etc.)
-    name = os.environ.get('RSTRNT_TASKNAME', '')
```
comps: This is not just Beaker native API, but also for OSCI, which would break with this removal.

lib/results.py
```diff
@@ -1,21 +1,17 @@
 """
 Functions and helpers for result management when working with
  - TMT (https://github.com/teemtee/tmt)
```
comps:
```suggestion
 - TMT (https://github.com/teemtee/tmt)
 - ATEX (https://github.com/RHSecurityCompliance/atex)
```
?

PR #489, opened 2025-11-13
Wait for osbuild compose build entry

comps:
This tries to work around some race conditions in osbuild composer where the blueprint appears only some time after a compose build has been started.

### DIFF
--- a/lib/osbuild.py
+++ b/lib/osbuild.py
@@ -128,9 +128,9 @@ class Compose:
 
     @classmethod
     def _wait_for_finished(cls, blueprint_name, sleep=1):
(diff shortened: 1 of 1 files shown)


comps:
The problem is somewhere else, this just causes a forever-wait for something that never appears.

So there likely isn't a race condition but something specific to `ccn_advanced` (9.4 at least), `bsi` (all RHEL-9) and `ism_o_top_secret` (all RHEL-10) that causes this. It's (likely) not a missing profile in some list inside osbuild, but another issue.

Maybe we need to manually reproduce this and investigate osbuild composer logs.

PR #507, opened 2025-12-18
Allow redirects when searching for url with install.img

matusmarhefka:
Redirects happen for some repositories (for example RDU points to BRQ).

There is one more commit which updates waivers and adds a waiver for new failing rule `file_permissions_boot_grub2` (issue reported in https://github.com/ComplianceAsCode/content/issues/14254).

### DIFF
--- a/conf/waivers/productization
+++ b/conf/waivers/productization
@@ -122,12 +122,9 @@
 /hardening/container/anaconda-ostree/.+/accounts_password_set_min_life_existing
 /hardening/container/anaconda-ostree/.+/accounts_password_set_warn_age_existing
     True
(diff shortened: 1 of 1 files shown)


jan-cerny (approved):
I created a virtual machine using this command:

```
./autocontest.sh install -c qemu:///system -n test_url -d rhel-8.10
```

Then, I run this test:

```
./autocontest.sh test -c qemu:///system -n test_url -t '/hardening/anaconda/stig'
```

With this patch I didn't get this error:
```
RuntimeError: did not find any install-capable repo amongst host repos
```

Thanks for your improvement!

PR #516, opened 2026-01-14
add `/plans/upstream` to be used by CaC/content CI

comps:


ggbecker:
it seems we also filter /static-checks/html-links (fails frequently due to temporary website availability issues)

https://github.com/ComplianceAsCode/content/blob/760b60746c89f9c593c900c87b6ec1359ec9b93e/tests/tmt/plans/contest.fmf#L120

PR #522, opened 2026-01-19
Condition `rhc-worker-playbook` installation by distro

comps:
This makes its usage more deterministic across RHEL / CentOS Stream rather than opportunistically installing it via `recommend` and using it based on whether it's installed.

The previous checking method could lead to other factors making `rhc-worker-playbook` not installed (ie. broken RPM in CentOS Stream), leading to inconsistent environments between executions.


comps:
Rebased, returned `scap-security-guide-rule-playbooks` to `recommend+`, and added a few other small fixes.

Issue #420, opened 2025-07-09
Switch over to standard TMT reporting, don't use `result:custom`


comps:
Since TMT gained support for "subresults", a more integrated way for a test to report fine-grained results, see ie. https://tmt.readthedocs.io/en/stable/spec/results.html , we no longer have to use the hacky `result:custom`.

This means we can gain (back) all the TMT features like test timeout tracking that [we implement ourselves](https://github.com/RHSecurityCompliance/contest/blob/682b8d6f9f5203102f1025970b766e6c083e431a/lib/runtest.py#L13-L46) by using regular `result:` (not specifying it in the `.fmf`) and reporting extra results as the `tmt-report-result` command, or generating the YAML bits under `subresults:`.

The command would be easier, but I don't think you can specify more than 1 log to it, for any given result. But that might be a reasonable tradeoff for TMT compatibility - Oculus (testing farm) doesn't seem to display files uploaded by subresults anyway, so we could just copy all logs to test `data` and be done.

PR #538, opened 2026-02-11
Switch to standard TMT reporting

matusmarhefka:
Resolves https://github.com/RHSecurityCompliance/contest/issues/420

Related tmt issue: https://github.com/teemtee/tmt/issues/4565


lib/results.py
```diff
@@ -33,8 +41,8 @@ def total(self):
 
 global_counts = Counter()
 
-# file for storing global logs across reboots (only for TMT and plain reporting)
-GLOBAL_LOGS_FILE = '_global_logs.json'
+# file that TMT reads for subresult entries (same as tmt-report-result writes to)
+TMT_REPORT_RESULTS_FILE = 'tmt-report-results.yaml'
```
comps: I would honestly just hardcode this inside `_write_tmt_subresult()`, because that's where I would look for how it works. The constant has no other use beyond that function and it will never change (is a TMT hardcoded implementation detail), so even exposing it as a module-level constant the user could theoretically change via
```python
from lib import results
results.TMT_REPORT_RESULTS_FILE = ...
```
has no use.

PR #581, opened 2026-03-30
Add a GH workflow for diffing FMF `adjust` sections

comps:


comps:
This was mostly a quick experiment and while it kind of works, there are some corner cases (ie. workflow not running after a `.fmf` modification was removed, leaving a stale comment) that I don't have time to fix.

Also, this is 100% AI-generated code and I'm not sure how I feel submitting this under my name - probably fine for a one-off offline script, but I wouldn't want to maintain the code.

PR #590, opened 2026-04-14
Add a test scanning hummingbird images

jan-cerny:
The new test `/scanning/hummingbird` will build Hummingbird product content and scan Hummingbird container image using `oscap-podman`.

The test consists of 2 tests `cis` and `stig` so that both existing profiles in the data stream are covered.

### DIFF
--- a/scanning/hummingbird/test.py
+++ b/scanning/hummingbird/test.py
@@ -1,0 +1,42 @@
+#!/usr/bin/python3
+
+import subprocess
+
+from lib import results, oscap, podman, util
+
+IMAGE = 'openjdk'
+profile = util.get_test_name().rpartition('/')[2]
+# Some profiles are expected to be applicable only to the FIPS variant of the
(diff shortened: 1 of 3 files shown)


scanning/hummingbird/test.py
```diff
@@ -0,0 +1,37 @@
...
+
+proc, lines = util.subprocess_stream(
+    [
+        'oscap-podman', image_id, 'xccdf', 'eval', '--profile', profile, '--progress',
+        '--report', 'report.html', '--results-arf', 'scan-arf.xml', ds_path,
+    ],
+    stderr=subprocess.STDOUT,
+)
+
+oscap.report_from_verbose(lines)
```
comps: This needs to be **inside** the context manager for `get_source_content()` - in some cases (working with a RPM), the content is built in a tmpdir that is removed after the context manager ends.

In that case `ds_path` would disappear.

matusmarhefka (approved):
LGTM

Issue #587, opened 2026-04-09
Remove `partitioning_mode = "raw"` customization in blueprints once OpenSCAP is updated to add it


matusmarhefka:
It was added in https://github.com/RHSecurityCompliance/contest/pull/586 and the following part of code can be removed once https://github.com/OpenSCAP/openscap/issues/2334 is resolved:
https://github.com/RHSecurityCompliance/contest/blob/183e64c055b6961a5e9c9cee78adf89fd1355615/lib/osbuild.py#L390-L420

PR #591, opened 2026-04-15
Change VM disk format from raw to qcow2 with 100GB virtual size

Arden97:
#### Description

Changed VM disk creation from 20GB raw format to 100GB qcow2 format with metadata preallocation

#### Rationale:

- GUI installations require more disk space
- `virt-install` cannot pass preallocation options to qemu-img
- `preallocation=metadata` provides near-raw performance with minimal overhead
- `/var/log/audit` size reduction hack in `translate_ssg_kickstart()` is no longer needed with 100GB virtual space
- Fix #559

#### Review hints

- `/hardening/oscap/stig` check was successfully run on the remote TF machine, disk pre-creation with appropriate 100 GB size was verified


lib/virt.py
```diff
@@ -414,14 +417,25 @@ def install_basic(
...
         disk_path = Path(f'{GUEST_IMG_DIR}/{self.name}.{disk_extension}')
 
+        # pre-create disk image with qemu-img to allow for preallocation=metadata
+        # which provides better performance for qcow2 images
+        util.log(f"pre-creating disk image {disk_path}")
+        qemu_img_cmd = ['qemu-img', 'create', '-f', disk_format]
+        if disk_format == 'qcow2':
+            qemu_img_cmd += ['-o', 'preallocation=metadata']
+        # 100 GB gives kickstarts plenty of space for partitions
+        qemu_img_cmd += [str(disk_path), '100G']
```
comps: `Path` to `str` conversion should AFAIK not be needed, `subprocess` takes `Path` arguments since Python 3.8 (and I'm pretty sure we use 3.9+ everywhere now?).

lib/virt.py
```diff
@@ -456,6 +470,7 @@ def install_basic(
             except Exception as e:
                 self.destroy()
                 self.undefine()
+                disk_path.unlink(missing_ok=True)
```
comps, line 472: This is done by `self.undefine(incl_storage=True)` if you want to do it. That way, UEFI nvram images (and snapshots, and checkpoints, etc.) also get removed.

comps (commented):
I left a few small comments.

Also, @matusmarhefka recently made https://github.com/RHSecurityCompliance/contest/pull/586 which introduces 30GB images and some hacks to make large partitions work. Should they also be increased to 100 GB, for consistency? Can any of the hacks be now removed, or are we waiting for OpenSCAP blueprint adjustments (https://github.com/RHSecurityCompliance/contest/issues/587)?

lib/virt.py
Arden97: Looking at `undefine` implementation:

```
    def undefine(self, incl_storage=False):
        if guest_domstate(self.name):
            storage = ['--remove-all-storage'] if incl_storage else []
            virsh(
                'undefine', self.name, '--nvram', '--snapshots-metadata',
                '--checkpoints-metadata', *storage, check=True,
            )
```

it runs only if domain is already defined. Don't we also want to handle situations, when `virt-install` fails at early stages (problems with libvirt, network, etc ...) ?

matusmarhefka:
> I left a few small comments.
> 
> Also, @matusmarhefka recently made #586 which introduces 30GB images and some hacks to make large partitions work. Should they also be increased to 100 GB, for consistency? Can any of the hacks be now removed, or are we waiting for OpenSCAP blueprint adjustments (#587)?

Actually yes, the disk image file generated by `composer-cli` is sparse so we can update
https://github.com/RHSecurityCompliance/contest/blob/183e64c055b6961a5e9c9cee78adf89fd1355615/lib/osbuild.py#L41-L44
to allocate even 100 GB:
```
QCOW2_IMAGE_SIZE_MIB = 100 * 1024
```
But the `partitioning_mode = "raw"` still needs to be set (until https://github.com/OpenSCAP/openscap/issues/2334 is resolved), otherwise the `/` partition won't be able to grow when IB composes an image and we might encounter `No space left on device` issues with some profiles like STIG.

lib/virt.py
comps: Yeah, fair point, we probably want to do both `incl_storage=True` and `disk_path.unlink(missing_ok=True)` if the intention is to remove storage.

Arden97:
PR was verified by custom productization pipeline

matusmarhefka (approved):
LGTM

lib/virt.py
```diff
@@ -414,14 +417,25 @@ def install_basic(
...
 
+        # pre-create disk image with qemu-img to allow for preallocation=metadata
+        # which provides better performance for qcow2 images
+        util.log(f"pre-creating disk image {disk_path}")
+        qemu_img_cmd = ['qemu-img', 'create', '-f', disk_format]
+        if disk_format == 'qcow2':
+            qemu_img_cmd += ['-o', 'preallocation=metadata']
+        # 100 GB gives kickstarts plenty of space for partitions
+        qemu_img_cmd += [disk_path, '100G']
+        subprocess.run(qemu_img_cmd, check=True, stderr=subprocess.PIPE)
```
matusmarhefka: Actually one last thing, please rather use here the `util.subprocess_run` instead of `subprocess.run` so the command is logged in the test log.

Arden97: @matusmarhefka done, also updated `subprocess.Popen` to `util.subprocess_Popen` for consistency

PR #592, opened 2026-04-15
Move sysctl forwarding waiver to permanent list for RHEL 8 GUI systems

ggbecker:
On RHEL 8 systems with GUI, libvirt (included in graphical-server-environment) sets /proc/sys/net/ipv4/ip_forward to 1 during boot, which also affects /proc/sys/net/ipv4/conf/all/forwarding. Libvirt requires forwarding enabled for VM networking to function properly. Since this is expected behavior and not a temporary productization issue, the waiver has been moved from the productization list to the permanent list.

Related: ComplianceAsCode/content#14532
Related: ComplianceAsCode/content#9316

### DIFF
--- a/conf/waivers/permanent
+++ b/conf/waivers/permanent
@@ -24,6 +24,28 @@
 /scanning/boot-errors/.+/.*systemd-journal-upload.*
     True
 
+# bz1825810 or maybe bz1929805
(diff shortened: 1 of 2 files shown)


matusmarhefka (approved):
LGTM

PR #593, opened 2026-04-16
Add retry wrapper for oscap per-rule scans

### LINKED ISSUE #548: Add `oscap` retries to `/per-rule`
The problem is that `oscap` still crashes a lot, we see notable amounts of
```
oscap exited unexpectedly with 139    (segfault)
```
amongst the `/per-rule` results (when testing various rules), which typically dooms the entire `/per-rule/*/$number` run, and a re-run catches some other rule segfaulting.

So maybe check if `oscap` as a scanning (not remediating) command exited with non-0 and if it did, try running it again.

Do that in a `for attempt in {1..5}` bash loop, so that we give it a few tries.

Arden97:
#### Description

- Add `oscap_scan_retry` function to repeat `per-rule` scan in case of unexpected exit codes (errors, segfaults, crashes, etc.)
- Function accepts a configurable `max_attempts` parameter (currently set to 5)

#### Rationale

- Fix #548

### DIFF
--- a/per-rule/runner.sh
+++ b/per-rule/runner.sh
@@ -57,7 +57,6 @@ function oscap_scan_retry {
     fi
 
     echo "$output"
-    return $rc
 }
 
 datastream=$thin_ds_dir/$rule.xml
--- a/per-rule/runner.sh
+++ b/per-rule/runner.sh
@@ -29,6 +29,31 @@ debug_arg=$5        # debug or nodebug
 
 function debug { [[ $debug_arg == debug ]]; }
 
+# Retry wrapper for oscap scans
+# Args: max_attempts oscap_args...
+# Retries on unexpected exit codes (errors, segfaults, crashes, etc.)
+# Only exit codes 0 and 2 are considered successful
+function oscap_scan_retry {
+    local max_attempts=$1
+    shift
+    local attempt rc output
+
+    for ((attempt=1; attempt<=max_attempts; attempt++)); do
+        output=$(oscap "$@"); rc=$?
+
+        # exit codes 0 and 2 are expected (success and rule failure)
+        if [[ $rc == 0 || $rc == 2 ]]; then
+            break
+        fi
+
+        # log the failure
+        echo "oscap attempt $attempt/$max_attempts failed with exit code $rc" >&2
+    done
+
+    echo "$output"
+    return $rc
+}
+
 datastream=$thin_ds_dir/$rule.xml
 playbook=$playbooks_dir/$rule.yml
 variables_file=$variables_dir/$rule/$test_name.$test_type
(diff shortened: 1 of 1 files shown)


per-rule/runner.sh
```diff
@@ -29,6 +29,31 @@ debug_arg=$5        # debug or nodebug
...
 function debug { [[ $debug_arg == debug ]]; }
 
+# Retry wrapper for oscap scans
+# Args: max_attempts oscap_args...
+# Retries on unexpected exit codes (errors, segfaults, crashes, etc.)
+# Only exit codes 0 and 2 are considered successful
+function oscap_scan_retry {
+    local max_attempts=$1
+    shift
+    local attempt rc output
```
comps: The current code is fine, but if you wanted to combine these, it can be done:

```suggestion
    local max_attempts=$1 attempt rc output
    shift
```

per-rule/runner.sh
```diff
@@ -29,6 +29,31 @@ debug_arg=$5        # debug or nodebug
...
+
+    for ((attempt=1; attempt<=max_attempts; attempt++)); do
+        output=$(oscap "$@"); rc=$?
```
comps: This will break in a `set -e` mode - if `oscap` exits with non-0, the entire shell exits, so `rc=$?` never gets to run.

The "usual hack" for this is to

```suggestion
        rc=0
        output=$(oscap "$@") || rc=$?
```

which causes the `||` branch to run on failure, and the next statement captures `$?`, with the overall expression having a return code of `0`, so `set -e` doesn't exit.

(Look around in the same file, it's used elsewhere.)

per-rule/runner.sh
```diff
@@ -29,6 +29,31 @@ debug_arg=$5        # debug or nodebug
...
+        # exit codes 0 and 2 are expected (success and rule failure)
+        if [[ $rc == 0 || $rc == 2 ]]; then
+            break
+        fi
+
+        # log the failure
+        echo "oscap attempt $attempt/$max_attempts failed with exit code $rc" >&2
+    done
+
+    echo "$output"
```
comps, line 59: I don't think we need to hide the output + display only the successful one. As long as we have the warning that oscap is being re-run due to crash (the log message above should probably mention something about oscap crashing and it being rerun), a reviewer can probably figure out what happened.

In fact - hiding the output could hide some more bugs (ie. oscap generating partial `report.html` + failing to overwrite it on the second run - hiding the first run would make debugging hard).

So I would just let `oscap` output without capture.

comps (dismissed):
Thanks for the simple implementation, I left a few minor notes, but overall it looks good!

per-rule/runner.sh
Arden97: Yeah, I didn't notice `set -x -e` at the beginning, fixed it in https://github.com/RHSecurityCompliance/contest/pull/593/commits/fbed35ba3c51ad25848d9ba14bf350bfd2ee996e

per-rule/runner.sh
comps: @Arden97 This is still waiting for your input - please either make the change or at least reply as to why you would prefer the current version, thanks!

comps:
Thanks, LGTM now. Running tests now.

PR #596, opened 2026-04-22
Add more testing to `plans-upstream`

comps:
It turns out we never configured the `/static-checks` to run in https://github.com/ComplianceAsCode/content/blob/07c4c5ac26ede5e3c132952209f50c074588f1bb/.packit.yaml#L26-L44 so that section of `plans/upstream-parallel.fmf` is dead.

So, instead of enabling it, I refactored it into `/other` and included `/static-checks` and some safe-enough `/scanning` tests too, as parallel plans.

The idea is that we only add a third `-other` Packit job (so we get `contest-oscap`, `contest-ansible`, `contest-other` for EACH CentOS Stream in CaC/content), but run multiple parallel plans within each Packit job.

In this case, each Ansible check-mode runs in parallel, and is a copy/paste of the above.

A second commit also refactors the original hardening oscap/ansible sections to reduce visual clutter - no behavior change, only syntax change.

---

~Do not merge yet, this uncovered some ansible playbook syntax issues that likely should be fixed first.~

### DIFF
--- a/plans/upstream-parallel.fmf
+++ b/plans/upstream-parallel.fmf
@@ -103,14 +103,66 @@ environment:
 /ansible/ccn_basic:
     discover+: {test: /hardening/host-os/ansible/ccn_basic$}
 
-/static-checks:
-    discover+:
-        test:
-          - /static-checks
-        exclude:
-          # frequently fails on transient issues, run it only in stabilization
-          - /static-checks/html-links
-          # not useful in automated runs that are not manually reviewed
-          - /static-checks/diff
+/other:
+    /static-checks:
+        discover+:
+            test:
+              - /static-checks
+            exclude:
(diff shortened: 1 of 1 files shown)


comps:
Actually, the `/other` plans don't touch current upstream configuration, because only oscap/ansible are used, and those still pass even with this new config.

So I think we can safely merge this, and a new PR against CaC/content will deal with any failures caused by `/other`.

comps:
Rebased, removed `/scanning/boot-errors` from `/other` since those tests use VMs and we don't have virtualization on generic TF systems via Packit.

Tested via https://github.com/comps/scap-content/pull/6 (see check for that PR) - `contest-other` won't run in current upstream.

matusmarhefka (approved):

PR #603, opened 2026-04-27
Update ansible.posix to 1.4.0 to be compatible with CentOS 10

macko1:
- CentOS 10 ships with Python 3.12.
- `ansible.posix` 1.3.0 depends on `distutils` which was deprecated in Python 3.12.
- Use the same ansible collections as equivalent RHEL versions (see which are included in the `rhc-worker-playbook`)

Tested locally with /static-checks/ansible/syntax-check - pass.
Tested in https://github.com/ComplianceAsCode/content/pull/14692 - pass.

### DIFF
--- a/lib/ansible.py
+++ b/lib/ansible.py
@@ -6,16 +6,18 @@ import subprocess
 
 from lib import util, results, versions
 
-# Versions pinned to match what rhc-worker-playbook bundles on RHEL, so that
-# CentOS/Fedora test results stay close to what customers run.
-#
+# Versions pinned to match what rhc-worker-playbook bundles on RHEL for the
+# given RHEL version, so that CentOS/Fedora test environments are as close
+# as possible to RHEL.
 # RHEL 8/9 (rhc-worker-playbook 0.1.x):
 #   https://gitlab.com/redhat/centos-stream/rpms/rhc-worker-playbook/-/blob/c9s/rhc-worker-playbook.spec
 # RHEL 10  (rhc-worker-playbook 0.2.x):
 #   https://github.com/RedHatInsights/rhc-worker-playbook/blob/main/ansible/meson.build
 ANSIBLE_GALAXY_COLLECTIONS = {
+    8: ['ansible.posix:1.3.0', 'community.general:4.4.0'],
+    9: ['ansible.posix:1.3.0', 'community.general:4.4.0'],
     10: ['ansible.posix:1.5.4', 'community.general:9.2.0'],
-    'default': ['ansible.posix:1.3.0', 'community.general:4.4.0'],
+    'default': ['ansible.posix', 'community.general'],
 }
 
 
@@ -29,14 +31,13 @@ def install_deps():
         os.environ['ANSIBLE_COLLECTIONS_PATH'] = \
             '/usr/share/rhc-worker-playbook/ansible/collections/ansible_collections/'
     else:
-        # On Fedora, CentOS, etc., fetch from Galaxy pinned to RHEL versions
-        # Default to RHEL 8/9 collections if no specific version is available
         collections = ANSIBLE_GALAXY_COLLECTIONS.get(
             versions.rhel.major, ANSIBLE_GALAXY_COLLECTIONS['default'],
         )
         util.subprocess_run(
-            ['ansible-galaxy', '-vvv', 'collection', 'install', '--force',
-             *collections],
+            (
+                'ansible-galaxy', '-vvv', 'collection', 'install', '--force', *collections,
+            ),
             check=True,
             stderr=subprocess.PIPE,
         )
--- a/lib/ansible.py
+++ b/lib/ansible.py
@@ -26,11 +26,19 @@ def install_deps():
     # Use ansible-galaxy when rhc-worker-playbook not available (Fedora, CentOS, etc.)
     else:
         for collection, version in RHC_WORKER_PLAYBOOK_COLLECTIONS.items():
+            # For CentOS Stream 10+, use ansible.posix 1.4.0 and newer,
+            # as it ships with Python 3.12 and it doesn't contain the distutils package,
+            # on which the ansible.posix 1.3.0 depends.
             if versions.rhel.is_centos():
-                # install the specific version to match rhc-worker-playbook versions
-                collection = f"{collection}:{version}"
+                if versions.rhel.major >= 10 and collection == 'ansible.posix':
+                    version = '1.4.0'
+
             util.subprocess_run(
-                ['ansible-galaxy', '-vvv', 'collection', 'install', collection],
+                ['ansible-galaxy',
+                '-vvv',
+                'collection', 'install',
+                f"{collection}:{version}",
+                '--force'],
                 check=True,
                 stderr=subprocess.PIPE,
             )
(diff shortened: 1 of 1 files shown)


lib/ansible.py
```diff
@@ -26,11 +26,19 @@ def install_deps():
     # Use ansible-galaxy when rhc-worker-playbook not available (Fedora, CentOS, etc.)
     else:
         for collection, version in RHC_WORKER_PLAYBOOK_COLLECTIONS.items():
+            # For CentOS Stream 10+, use ansible.posix 1.4.0 and newer,
+            # as it ships with Python 3.12 and it doesn't contain the distutils package,
+            # on which the ansible.posix 1.3.0 depends.
             if versions.rhel.is_centos():
-                # install the specific version to match rhc-worker-playbook versions
-                collection = f"{collection}:{version}"
+                if versions.rhel.major >= 10 and collection == 'ansible.posix':
```
comps: `versions.rhel >= 10` should be enough, you don't need to access `major`.

Also, consider combining it with the `if` above.

comps:
Why the extra line break removals in Markdown files?

macko1:
@comps starting using https://github.com/pre-commit/pre-commit-hooks#trailing-whitespace so I figured it wouldn't hurt. No line breaks removed. 

Should I revert?

lib/ansible.py
macko1: I wanted to do as little changes as possible to the pipeline. To cover only the failing scenario.

We can test for other versions, too, as the collection versions used are quite ancient. 

Would you find it useful to get rid of the conditions and update the collections for all platforms?

comps:
Yes, please revert those since it changes the formatting. Markdown uses either two trailing whitespaces or `\` at line end to denote a line break.

Using `\` might be better over `  `, but that's probably for another PR or commit, when done intentionally.

lib/ansible.py
comps: I was just commenting on the code style, not on doing behavioral changes.

Generally, we'd want to follow https://docs.astral.sh/ruff/rules/collapsible-if/, though we don't let ruff enforce this, since it lacks the nuance of where it's acceptable and where it would complicate things.

```python
if versions.rhel.is_centos():
    if versions.rhel.major >= 10 and collection == 'ansible.posix':
        ...
```
is the same as 
```python
if versions.rhel.is_centos() and versions.rhel >= 10 and collection == 'ansible.posix':
    ...
```
but since it would exceed the max line length, keeping them separate is probably nicer here.

macko1:
@comps: Ah, didn't know that, and yes using trailing whitespace to represent anything is not ideal. Reverting.

I will amend the commit and rebase it on main.

lib/ansible.py
comps: Actually, sorry I didn't realize this sooner, since Contest doesn't support Fedora, we're dealing with just RHEL or CentOS Stream here, so we can probably drop the whole `if versions.rhel.is_centos():` below, incl. the version override, and just define it here, since the constants are already outlined above the code.

```python
    "ansible.posix": "1.3.0" if versions.rhel < 10 else "1.4.0",
```

WDYT?

macko1: I think the main reason for pinning the versions is to be as close to RHEL as possible. Is my assumption correct @matusmarhefka @ggbecker ?

I checked the current versions at
```
# RHEL 8/9 (rhc-worker-playbook 0.1.x):
# https://gitlab.com/redhat/centos-stream/rpms/rhc-worker-playbook/-/blob/c9s/rhc-worker-playbook.spec
# RHEL 10  (rhc-worker-playbook 0.2.x):
# https://github.com/RedHatInsights/rhc-worker-playbook/blob/main/ansible/meson.build
```
and updated the code.


I will test this in our CI.

matusmarhefka: Yes, we want to use the same versions as in `rhc-worker-playbook` in case it is not available/possible to install.

macko1: The tests are passing with the currently-pinned versions in https://github.com/ComplianceAsCode/content/pull/14692 

If you don't have any objections, I believe we can merge and close this. Thanks for the comments, guys.
@matusmarhefka @comps

matusmarhefka:
Current changes look good to me, but I will let merging of this PR for @comps as a primary reviewer.

lib/ansible.py
```diff
@@ -6,34 +6,40 @@
...
+# Versions pinned to match what rhc-worker-playbook bundles on RHEL, so that
+# CentOS/Fedora test results stay close to what customers run.
+#
+# RHEL 8/9 (rhc-worker-playbook 0.1.x):
+#   https://gitlab.com/redhat/centos-stream/rpms/rhc-worker-playbook/-/blob/c9s/rhc-worker-playbook.spec
+# RHEL 10  (rhc-worker-playbook 0.2.x):
+#   https://github.com/RedHatInsights/rhc-worker-playbook/blob/main/ansible/meson.build
+ANSIBLE_GALAXY_COLLECTIONS = {
+    10: ['ansible.posix:1.5.4', 'community.general:9.2.0'],
+    'default': ['ansible.posix:1.3.0', 'community.general:4.4.0'],
```
comps: Why not make it future-proof?

```suggestion
    8: ['ansible.posix:1.3.0', 'community.general:4.4.0'],
    9: ['ansible.posix:1.3.0', 'community.general:4.4.0'],
    10: ['ansible.posix:1.5.4', 'community.general:9.2.0'],
    'default': ['ansible.posix', 'community.general'],
```

This would use latest version (I think) on RHEL-11 / Fedora (if we ever support it), rather than using the oldest collections for unspecified releases.

lib/ansible.py
```diff
@@ -6,34 +6,40 @@
...
-                stderr=subprocess.PIPE,
-            )
+        # On Fedora, CentOS, etc., fetch from Galaxy pinned to RHEL versions
+        # Default to RHEL 8/9 collections if no specific version is available
+        collections = ANSIBLE_GALAXY_COLLECTIONS.get(
+            versions.rhel.major, ANSIBLE_GALAXY_COLLECTIONS['default'],
+        )
+        util.subprocess_run(
+            ['ansible-galaxy', '-vvv', 'collection', 'install', '--force',
+             *collections],
```
comps: I don't think this matches https://github.com/RHSecurityCompliance/contest/blob/main/docs/STYLE.md#close-blocks-on-the-same-indent-level , though it's interesting that `flake8-picky-parenthesis` didn't pick this up. Maybe because it's `[ ]` and not `( )`. 🫤 

```suggestion
            [
                'ansible-galaxy', '-vvv', 'collection', 'install', '--force', *collections,
            ],
```

macko1: Seems like `flake8-picky-parenthesis`  plugin doesn't cover this case, does not matter if it is a tuple or a list.

```
cat <<'EOF' > /tmp/test_paren.py
x = func(
    ('a', 'b',
     'c'),
    y=1,
)
EOF
python3 -m flake8 --isolated --require-plugins flake8-picky-parentheses --select PAR1 /tmp/test_paren.py
```
Returns 0.

macko1: Fixed.

lib/ansible.py
macko1: Yes, I think we can use this. 

Wasn't thinking about future-proofing, as this requires manual updates with every update to the collections referenced in the `rhc-worker-playbook`, anyway.

Fixed.

PR #606, opened 2026-04-30
Extend Hummingbird scanning test

jan-cerny:
This commit changes the `/scanning/hummingbird` tests with these 2 major improvements:

1. The test has been split into 2 tests:
   1. scanning using oscap-podman
   2. scanning using the openscap hummingbird container image
2. The test now tests more profile and image variants:
   1. normal images are scanned with CIS profiles
   2. FIPS images are scanned with both CIS and STIG profiles

The reason for the improvement 1 is that while users can use oscap-podman for scanning, the hummingbird users will be guided by documentation to use the scanner and content shipped in the `quay.io/hummingbird/openscap` container image.

The reason for the improvement 2 CIS profile should be applicable also on FIPS variants of images. Remember that hummingbird provides images in 2 basic variants:
1. normal (eg. `:latest`) - compliant only with CIS
2. FIPS (eg. `:latest-fips`) - compliant with CIS and STIG

Related to: https://github.com/ComplianceAsCode/content/pull/14683

### DIFF
--- a/scanning/hummingbird/image/test.py
+++ b/scanning/hummingbird/image/test.py
@@ -4,8 +4,12 @@ import shutil
 import subprocess
 import tempfile
 
-from lib import results, metadata, oscap, podman, util
 from pathlib import Path
+
+from lib import results, metadata, oscap, podman, util
+
+SCANNER_IMAGE = 'quay.io/hummingbird/openscap:latest'
+podman.podman('pull', SCANNER_IMAGE)
 
 IMAGE = 'openjdk'
 profile = util.get_test_name().rpartition('/')[2]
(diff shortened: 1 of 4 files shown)


scanning/hummingbird/image/test.py
```diff
@@ -31,20 +35,26 @@
         shutil.copy(ds_path, tempdir)
```
comps: I think this whole tempdir handling is unnecessary - all tests are already guaranteed to run inside a temporary directory, see https://github.com/RHSecurityCompliance/contest/blob/1af7fc729f1a2abf2667e553cf80cb8436802723/lib/runtest.py#L26-L40

So you can just copy the datastream to `.`:

```python
shutil.copy(ds_path, '.')
```

scanning/hummingbird/image/test.py
```diff
@@ -31,20 +35,26 @@
...
         )
         oscap.report_from_verbose(lines)
         if proc.returncode not in [0, 2]:
             raise RuntimeError("oscap failed unexpectedly")
-        shutil.copy(tempdir + '/scan-arf.xml', Path.cwd())
-        shutil.copy(tempdir + '/report.html', Path.cwd())
+        artifacts = ['report.html', 'scan-arf.xml']
+        tempdir_path = Path(tempdir)
+        for artifact in artifacts:
+            shutil.copy(tempdir_path / artifact, Path.cwd())
```
comps: Similarly here, you don't need to copy anything - just add logs from the current directory using relative paths, no need for absolute `Path.cwd()`.

```python
results.add_log('report.html', 'scan-arf.xml')
```

and, at the end, just call

```python
results.report_and_exit()
```

@matusmarhefka This should work, right? Thanks!

jan-cerny:
I have removed creatingof temporary directory

scanning/hummingbird/image/test.py
matusmarhefka: Yes, that should work.

scanning/hummingbird/image/test.py
```diff
@@ -0,0 +1,55 @@
...
+            '--report', '/ssg/report.html',
+            '/ssg/ssg-hummingbird-ds.xml',
+        ],
+        stderr=subprocess.STDOUT,
+    )
+    oscap.report_from_verbose(lines)
+    if proc.returncode not in [0, 2]:
+        raise RuntimeError("oscap failed unexpectedly")
+
+results.add_log('report.html', 'scan-arf.xml')
```
matusmarhefka: This is not needed now as `results.add_log` got moved out of the previous block and you are calling `results.report_and_exit()` right below it. So we can spare one unnecessary line and upload logs directly in `results.report_and_exit()`:
```
results.report_and_exit(logs=['report.html', 'scan-arf.xml'])
```

jan-cerny:
I have changed report and exit

matusmarhefka (approved):

PR #607, opened 2026-04-30
limiting waiver of issue 14581 to RHEL <= 9.7

vojtapolasek:
because since 9.8 it is fixed


matusmarhefka:
@vojtapolasek I see the linked issue is closed. If we plan to keep these waivers they should be moved to the permanent file instead of productization.

vojtapolasek:
I moved the waiver to permanent.

PR #608, opened 2026-05-07
Disable all automatic runs of the /static-checks/audit-sample-rules test

vojtapolasek:
We want to keep this test, but we want to be able to run it only when needed.

### DIFF
--- a/plans/daily.fmf
+++ b/plans/daily.fmf
@@ -39,5 +39,7 @@ discover:
       - /static-checks/html-links
       # not useful in automated runs that are not manually reviewed
       - /static-checks/diff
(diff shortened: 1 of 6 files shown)


comps:
I would suggest to remove the test completely - if we re-introduce this as a workflow (Jira) based process, it should probably be using `diff` or something like that, not rely on Contest.

@matusmarhefka WDYT?

In either case, I'd remove the waiver for it:
```
# https://github.com/ComplianceAsCode/content/issues/14580
/static-checks/audit-sample-rules/audit_ospp_general
    rhel in [8, 9]
```

vojtapolasek:
OK, I will create a new PR.

PR #622, opened 2026-05-20
Extend waiver for service_kdump_disabled

jan-cerny:
The pre-release stabilization of the 0.1.81 release has revealed that the issue about service_kdump_disabled affects all 3 ISM profiles, not only the top secret profile, therefore we will extend the waiver to all 3 ISM profiles.

### DIFF
--- a/conf/waivers/productization
+++ b/conf/waivers/productization
@@ -34,7 +34,7 @@
 
 # kdump being forcibly enabled somewhere
 # https://github.com/ComplianceAsCode/content/issues/12832
-/hardening/kickstart(/with-gui|/uefi)?/(hipaa|stig|stig_gui|ism_o|ism_o_secret|ism_o_top_secret|e8)/service_kdump_disabled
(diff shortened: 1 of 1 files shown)


conf/waivers/productization
```diff
@@ -34,7 +34,7 @@
 
 # kdump being forcibly enabled somewhere
 # https://github.com/ComplianceAsCode/content/issues/12832
-/hardening/kickstart(/with-gui|/uefi)?/(hipaa|stig|stig_gui|ism_o_top_secret|e8)/service_kdump_disabled
+/hardening/kickstart(/with-gui|/uefi)?/(hipaa|stig|stig_gui|ism_o|ism_o_secret|ism_o_top_secret|e8)/service_kdump_disabled
```
comps: Since it seems clear this impacts all profiles having the rule, let's just wildcard it:


```suggestion
/hardening/kickstart/.+/service_kdump_disabled
```

PR #633, opened 2026-06-15
Use ATEX file upload appending for line-by-line log streaming

### LINKED ISSUE #521: Use ATEX file upload appending for logs
Now that https://github.com/RHSecurityCompliance/atex supports appending to previously-uploaded files (simply by specifying the same file name again in another `partial=True` call), we probably should use it for some key log files which we process line-by-line.

For example
* `lib/ansible.py` has `ansible-playbook` line-by-line parsing in https://github.com/RHSecurityCompliance/contest/blob/7134bf25c48d2272553dbe0c96ecb0c7411ff476/lib/ansible.py#L42-L48 and it could add a `to_atex='ansible-playbook.log'` argument, routing the line-by-line output to something like `report_atex(status='error', logs=['ansible-playbook.log'], partial=True)` or a more efficient function that takes a `data=` argument and not `logs=` as temporary files
* `lib/oscap.py` also has line-by-line `oscap` output printing in https://github.com/RHSecurityCompliance/contest/blob/7134bf25c48d2272553dbe0c96ecb0c7411ff476/lib/oscap.py#L211-L223 and could be similarly modified, to output everything (and not just non-progress lines) to `oscap.log`, avoiding the output in `output.txt` (meaning we could make `output.txt` uncompressed, useful in upstream)
* `lib/virt.py` does print out all `virt-install` output as https://github.com/RHSecurityCompliance/contest/blob/7134bf25c48d2272553dbe0c96ecb0c7411ff476/lib/virt.py#L444-L446 which was an issue in the past (when we had verbose install output) - if we routed that to a separate compressed `virt-install.log`, we could re-enable the verbose output, which is necessary for debugging most issues (currently we work around it by manually reproducing the issue and investigating `/tmp/*.log`)

matusmarhefka:
Add `atex_upload_log_data()` to `lib/results.py`, which streams raw data
to a named ATEX log file using partial result appending (same filename
in multiple `partial=True` calls causes ATEX to append to the file).

This allows `ansible-playbook`, `oscap` and `virt-install` output to be
uploaded to ATEX incrementally as it is produced, providing crash-safety
(if the test dies mid-execution, ATEX already has all output up to that
point) and keeping `output.txt` smaller.

Add `register_log()` which pre-registers a log file so it is preserved
even if the test errors out. For TMT, it writes directly into
`TMT_TEST_DATA` and registers the name in `TMT_TEST_SUBMITTED_FILES`.
For ATEX, streaming already provides crash-safety so the original path
is returned as-is. Library functions call `register_log()` internally, so
callers no longer need to call `add_log()` for streamed log files.

Consolidate ATEX protocol into shared helpers: `_atex_write_all()` retries
partial writes and `EINTR`, `_atex_send()` sends the JSON header then either
inline data (per-line streaming) or log files read in 64 KiB chunks.

Changes to library modules:
- `lib/results.py`: `add _atex_write_all()`, `_atex_send()`, `_get_atex_fd()`,
  `atex_upload_log_data()` and `register_log()`; modify `add_log()` and
  `report_atex()` to skip re-uploading already-streamed logs
- `lib/ansible.py`: `report_from_output()` always writes to a log file
  (default `ansible-playbook.log`) and streams each line to ATEX;
  the optional `sys.stdout` fallback is removed
- `lib/oscap.py`: `report_from_verbose()` always writes to a log file
  (default `oscap.log`) and streams each line to ATEX; the old
  `rules_from_verbose()` stdout printing is replaced by inline
  `rule_from_verbose()` matching in the same write loop
- `lib/virt.py`: `install_basic()` writes `virt-install` output to a log
  file and streams it to ATEX instead of writing to stdout

Changes to tests:
- All callers of `ansible.report_from_output()` and `oscap.report_from_verbose()`
  no longer call `results.add_log()` for streamed log files, as `register_log()` handles it internally

Resolves: #521

### DIFF
--- a/lib/ansible.py
+++ b/lib/ansible.py
@@ -1,7 +1,5 @@
 import os
-import sys
 import re
-import contextlib
 import subprocess
 
 from lib import util, results, versions
@@ -43,13 +41,17 @@ def install_deps():
         )
 
 
-def report_from_output(lines, to_file=None, failure='fail'):
+def report_from_output(lines, to_file='ansible-playbook.log', failure='fail'):
     """
     Process 'ansible-playbook' output, hide useless info, and report important
     info.
 
-    If 'to_file' was specified as a file path, redirect ansible-playbook
-    outputs to it, instead of leaving them on the console.
+    All ansible-playbook output is written to 'to_file' (and streamed to ATEX
+    when available). The log file is pre-registered via register_log() so it
+    is preserved even if the test errors out (similar to ATEX crash-safety).
+    Callers should not pass 'to_file' to results.add_log() or
+    results.report_and_exit(logs=...) as it is already registered with
+    the main test result.
 
     The 'failure' argument dictates what status to use for failing playbook
     results. The default 'fail' is sensible for uses where the playbook is
@@ -62,19 +64,16 @@ def report_from_output(lines, to_file=None, failure='fail'):
     """
     failed = False
     task = '<unknown task>'
+    log_path = results.register_log(to_file)
 
-    with contextlib.ExitStack() as stack:
-        if to_file:
-            out_file = stack.enter_context(open(to_file, 'w'))
-        else:
-            out_file = sys.stdout
-
+    with open(log_path, 'w') as out_file:
         for line in lines:
             # shorten facts
             m = re.match(r'(ok: \[.+\] =>) {"ansible_facts": ', line)
             if m:
                 line = f'{m.group(1)} (Redacted by Contest)'
 
+            results.atex_upload_log_data(to_file, f'{line}\n')
             out_file.write(f'{line}\n')
             out_file.flush()
 
--- a/lib/oscap.py
+++ b/lib/oscap.py
@@ -1,4 +1,3 @@
-import sys
 import re
 import enum
 import contextlib
@@ -208,24 +207,16 @@ def rule_from_verbose(line):
         return None
 
 
-def rules_from_verbose(lines):
-    """
-    Yield (rulename, status) from oscap info verbose output lines.
-    """
-    for line in lines:
-        # oscap xccdf eval --progress rule name and status
-        match = rule_from_verbose(line)
-        if match:
-            yield match
-        else:
-            # print out unrelated lines
-            sys.stdout.write(f'{line}\n')
-            sys.stdout.flush()
-
-
-def report_from_verbose(lines):
+def report_from_verbose(lines, to_file='oscap.log'):
     """
     Report results from oscap output.
+
+    All oscap output lines are written to 'to_file' (and streamed to ATEX
+    when available) instead of being printed to stdout. The log file is
+    pre-registered via register_log() so it is preserved even if the test
+    errors out (similar to ATEX crash-safety). Callers should not pass
+    'to_file' to results.add_log() or results.report_and_exit(logs=...)
+    as it is already registered with the main test result.
 
     Note that this expects 'oscap xccdf eval' to be run:
       - with --progress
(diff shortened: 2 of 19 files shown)


lib/ansible.py
```diff
@@ -49,7 +50,9 @@ def report_from_output(lines, to_file=None, failure='fail'):
     info.
 
     If 'to_file' was specified as a file path, redirect ansible-playbook
-    outputs to it, instead of leaving them on the console.
+    outputs to it, instead of leaving them on the console. Additionally,
+    when ATEX is available, the output is streamed to ATEX under the file's
+    basename.
```
comps: Why basename? ATEX supports full sub-paths in file names, so if somebody uses `to_file='foo/bar.log'` it should AFAIK work just fine.

lib/oscap.py
```diff
@@ -213,43 +213,57 @@ def rules_from_verbose(lines):
...
-            status = 'error'
+            out_file = sys.stdout
+
+        def tee(lines):
+            for line in lines:
+                if atex_name:
+                    results.atex_upload_log_data(atex_name, f'{line}\n')
+                out_file.write(f'{line}\n')
+                out_file.flush()
+                yield line
```
comps: Why do we write to console here via `tee()`, but not for ansible (above) where the writing-to-stdout was removed completely?

lib/results.py
```diff
@@ -94,6 +99,61 @@ def _tmt_file_submit(filepath):
...
 
 
+_atex_fd = None
+
+
+def _get_atex_fd():
+    global _atex_fd
+    if _atex_fd is None:
+        _atex_fd = int(os.environ['ATEX_TEST_CONTROL'])
+    return _atex_fd
```
comps: Why the caching? Is getting a variable from `os.environ` and converting it to an integer an expensive operation?

comps (commented):
I glanced a bit through the code, provided a few comments inline.

lib/results.py
matusmarhefka: I did it like this because the variable might be retrieved and converted to an integer thousands of times, for each line of file which is streamed do ATEX (including `virt-install` debug log from VM installation).

comps:
```
$ ABC=123 python3 -m timeit -s 'import os' 'int(os.environ["ABC"])'
2000000 loops, best of 5: 152 nsec per loop

$ ABC=123 python3 -m timeit -s $'import os\nglobals()["_cache"] = None\ndef func():\n  global _cache\n  if _cache is None:\n    _cache = int(os.environ["ABC"])\n  return _cache\n' 'func()'
20000000 loops, best of 5: 15.2 nsec per loop
```
Okay, the cache is about 10x faster, even though both are already really fast. So it's your call. :slightly_smiling_face:

lib/results.py
```diff
@@ -260,12 +369,16 @@ def add_log(*logs):
...
+        for log in logs:
+            log = Path(log)
+            if str(log) in _streamed_atex_logs:
+                continue
+            with open(log, 'rb') as f:
+                while True:
+                    chunk = f.read(65536)
+                    if not chunk:
+                        break
+                    atex_upload_log_data(str(log), chunk)
```
comps:
```suggestion
                while chunk := f.read(65536):
                    atex_upload_log_data(str(log), chunk)
```
?

lib/results.py
```diff
@@ -241,13 +297,66 @@ def report(status, name=None, note=None, logs=None):
...
+    filename in multiple partial=True calls appends to the file.
+    Outside ATEX, this is a no-op.
+
+    'data' is appended as-is (bytes or str, encoded to UTF-8 if str).
+    The caller is responsible for including any line terminators.
+    """
+    if not have_atex_api():
+        return
+    if isinstance(data, str):
+        data = data.encode()
```
comps, line 299:
```suggestion
        data = data.encode(errors='replace')
```
?

To guard against non-unicode non-ascii binary garbage in log files.

comps: Actually, nevermind, this is encoding, not decoding.

lib/results.py
```diff
@@ -80,18 +85,80 @@ def _write_tmt_subresult(subresult):
...
+    """
+    view = memoryview(data)
+    while view:
+        try:
+            nbytes = os.write(fd, view)
+        except InterruptedError:
+            continue
+        if nbytes == 0:
+            raise OSError("unexpected zero-length write on ATEX control fd")
+        view = view[nbytes:]
```
comps: Is this needed on modern Python versions that don't propagate `EINTR`? I think Python 3.5+ takes care of this.

PR #634, opened 2026-06-24
per-rule tests: detect and test rules with SCE content

vojtapolasek:
Heavily based on recommendations from https://github.com/ComplianceAsCode/content/pull/13758#issuecomment-3155084027

Test results of OVAL rules do not change.
If the rule has SCE, the test result appends "_sce" to the rule name.
In case the rule has ONLY SCE, the OVAL test is marked as skipped.

### DIFF
--- a/lib/oscap.py
+++ b/lib/oscap.py
@@ -14,6 +14,15 @@ FixType = enum.Flag(
     ['bash', 'ansible', 'anaconda', 'kickstart', 'blueprint', 'bootc'],
 )
 
+_FIX_SYSTEM_MAP = {
+    'urn:xccdf:fix:script:sh': FixType.bash,
+    'urn:xccdf:fix:script:ansible': FixType.ansible,
+    'urn:redhat:anaconda:pre': FixType.anaconda,
+    'urn:xccdf:fix:script:kickstart': FixType.kickstart,
+    'urn:redhat:osbuild:blueprint': FixType.blueprint,
+    'urn:xccdf:fix:script:bootc': FixType.bootc,
+}
+
 
 def parse_xml(path):
     """
@@ -133,20 +142,10 @@ class Datastream:
 
             # fixes / remediations
             elif frames[-2:] == ['Rule', 'fix']:
-                system = elements[-1].get('system')
-                for_rule = elements[-1].get('id')
-                if system == 'urn:xccdf:fix:script:sh':
-                    self.rules[for_rule].fixes |= FixType.bash
-                elif system == 'urn:xccdf:fix:script:ansible':
-                    self.rules[for_rule].fixes |= FixType.ansible
-                elif system == 'urn:redhat:anaconda:pre':
-                    self.rules[for_rule].fixes |= FixType.anaconda
-                elif system == 'urn:xccdf:fix:script:kickstart':
-                    self.rules[for_rule].fixes |= FixType.kickstart
-                elif system == 'urn:redhat:osbuild:blueprint':
-                    self.rules[for_rule].fixes |= FixType.blueprint
-                elif system == 'urn:xccdf:fix:script:bootc':
-                    self.rules[for_rule].fixes |= FixType.bootc
+                fix_type = _FIX_SYSTEM_MAP.get(elements[-1].get('system'))
+                if fix_type:
+                    for_rule = elements[-1].get('id')
+                    self.rules[for_rule].fixes |= fix_type
 
         # "convert" to regular dict, make external logic get KeyError
         # on bad profile or rule name
--- a/lib/oscap.py
+++ b/lib/oscap.py
@@ -14,6 +14,8 @@ FixType = enum.Flag(
     ['bash', 'ansible', 'anaconda', 'kickstart', 'blueprint', 'bootc'],
 )
 
+_OVAL_SYSTEM = 'http://oval.mitre.org/XMLSchema/oval-definitions-5'
+
 _FIX_SYSTEM_MAP = {
     'urn:xccdf:fix:script:sh': FixType.bash,
     'urn:xccdf:fix:script:ansible': FixType.ansible,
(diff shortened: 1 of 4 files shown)


lib/oscap.py
```diff
@@ -14,6 +14,15 @@
...
 
+_FIX_SYSTEM_MAP = {
+    'urn:xccdf:fix:script:sh': FixType.bash,
+    'urn:xccdf:fix:script:ansible': FixType.ansible,
+    'urn:redhat:anaconda:pre': FixType.anaconda,
+    'urn:xccdf:fix:script:kickstart': FixType.kickstart,
+    'urn:redhat:osbuild:blueprint': FixType.blueprint,
+    'urn:xccdf:fix:script:bootc': FixType.bootc,
+}
+
```
comps, line 25: I don't think this is necessary, Claude refactoring this to avoid the BS logic of PLR0915 instead of ignoring it in `ruff.toml` is IMHO wrong.

It splits the algorithm to multiple places for no real benefit - in fact, once our supported Python version increases to 3.10+, we could use `match`/`case`:

```python
  elif frames[-2:] == ['Rule', 'fix']:
      system = elements[-1].get('system')
      for_rule = elements[-1].get('id')
      match system:
          case 'urn:xccdf:fix:script:sh':
              self.rules[for_rule].fixes |= FixType.bash
          case 'urn:xccdf:fix:script:ansible':
              self.rules[for_rule].fixes |= FixType.ansible
          case 'urn:redhat:anaconda:pre':
              self.rules[for_rule].fixes |= FixType.anaconda
          case 'urn:xccdf:fix:script:kickstart':
              self.rules[for_rule].fixes |= FixType.kickstart
          case 'urn:redhat:osbuild:blueprint':
              self.rules[for_rule].fixes |= FixType.blueprint
          case 'urn:xccdf:fix:script:bootc':
              self.rules[for_rule].fixes |= FixType.bootc
```

That said, the rewrite is not strictly wrong, up to @matusmarhefka (as the head maintainer) to decide.

lib/oscap.py
```diff
@@ -140,12 +142,23 @@ def make_rule():
...
+                if frames[-1] == 'fix':
+                    fix_type = _FIX_SYSTEM_MAP.get(system)
+                    if fix_type:
+                        for_rule = elements[-1].get('id')
+                        self.rules[for_rule].fixes |= fix_type
+                elif frames[-1] == 'check':
+                    rule_id = elements[-2].get('id')
+                    rule_id = rule_id.removeprefix(
+                        'xccdf_org.ssgproject.content_rule_',
+                    )
```
comps, line 157: This doesn't need to be multiline, the project allows up to 99 characters on a single line, according to `ruff.toml`, `line-length = 99`.

lib/oscap.py
```diff
@@ -140,12 +142,23 @@ def make_rule():
...
+                    if system == 'http://open-scap.org/page/SCE':
+                        self.rules[rule_id].has_sce = True
+                    elif system == _OVAL_SYSTEM:
+                        self.rules[rule_id].has_oval = True
```
comps, line 161: If you hardcoded `http://open-scap.org/page/SCE`, why not also `http://oval.mitre.org/XMLSchema/oval-definitions-5`? Why is one a constant and one hardcoded?

per-rule/setup.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+dnf install -y openscap-engine-sce
```
comps, line 5: This belongs to the VM installation section, in `test.py`, there's already

```python
    # prepare a kickstart for the VM
    ks = virt.Kickstart()
    ks.packages += ['rsync', 'xmlstarlet', 'ansible-core']
```

You can simply add the package there, assuming it's always available across all supported RHEL-8/9/10 (and CentOS Stream) versions.

per-rule/test.py
```diff
@@ -62,6 +63,8 @@ def unit_tests_from_rules(built_tests_dir, rules):
                 if full.rule in remediation_excludes:
                     continue
             yield full
+            if rule in ds.rules and ds.rules[rule].has_sce:
```
comps, line 66: Why the `if rule in ds.rules`? What is it guarding against? When is `rule` not in `ds.rules`? We probably shouldn't ever pass invalid rule names to `unit_tests_from_rules()`.

per-rule/test.py
```diff
@@ -194,6 +197,13 @@ def unit_tests_from_rules(built_tests_dir, rules):
     if test.remediation == 'ansible' and fix_type != 'ansible':
         continue
 
+    # SCE-only rule with no OVAL check - skip the non-SCE test
+    if test.check != 'sce' and test.rule in ds.rules:
```
comps, line 201: Again, why the `test.rule in ds.rules` ?

per-rule/test.py
```diff
@@ -194,6 +197,13 @@ def unit_tests_from_rules(built_tests_dir, rules):
...
+        rule_info = ds.rules[test.rule]
+        if rule_info.has_sce and not rule_info.has_oval:
+            results.report('skip', format_test(test), 'rule has no OVAL check')
```
comps, line 204: Why is this a skip? If a rule has SCE only, isn't that actually wanted for some complex rules that can't be covered by OVAL?

per-rule/test.py
```diff
@@ -62,6 +63,8 @@ def unit_tests_from_rules(built_tests_dir, rules):
                 if full.rule in remediation_excludes:
                     continue
             yield full
+            if rule in ds.rules and ds.rules[rule].has_sce:
+                yield full._replace(check='sce')
```
comps, line 67: Also, why leave `check` empty for OVAL? Why not simply

```python
if ds.rules[rule].has_oval:
    yield full._replace(check='oval')
if ds.rules[rule].has_sce:
    yield full._replace(check='sce')
```

?

That would ensure `check` is always set.

per-rule/test.py
```diff
@@ -205,7 +215,8 @@ def unit_tests_from_rules(built_tests_dir, rules):
     # try a first run, if it passes, report it without extra overhead
     with g.snapshotted():
         proc = g.ssh(
-            './runner.sh', test.rule, test.test, pass_fail, remediation_type, 'nodebug',
+            './runner.sh', test.rule, test.test, pass_fail,
+            remediation_type, 'nodebug', test.check or 'oval',
```
comps, line 219: This is again hardcoding `oval` instead of just passing `test.check` as-is, because the code in `unit_tests_from_rules()` returns empty `test.check`, which it probably shouldn't.

comps (commented):
I've left a few comments, hopefully that helps.

PR #635, opened 2026-06-26
Remove waivers for RHEL 10 STIG

jan-cerny:
The RHEL 10 STIG profile is now following the official DISA STIG V1R1 benchmark. That has been merged upstream in
https://github.com/ComplianceAsCode/content/pull/14826 Consequently, the rules in the profile now have STIG IDs and references.  In 2026-06-26 daily productization, the tests checking these IDs or references started passing and are displayed as "waived pass".  That means we can remove waivers for these tests.


vojtapolasek (approved):
lgtm

PR #636, opened 2026-06-26
per-rule tests: detect and test rules with SCE content

vojtapolasek:
Heavily based on recommendations from https://github.com/ComplianceAsCode/content/pull/13758#issuecomment-3155084027

Test results of OVAL rules do not change.
If the rule has SCE, the test result prepends  "/sce" before the rule name.

### DIFF
--- a/per-rule/test.py
+++ b/per-rule/test.py
@@ -31,7 +31,10 @@ remediation_excludes = set(remediation.excludes())
 
 def format_test(test):
     pass_fail = 'pass' if test.is_pass else 'fail'
-    return f'{test.rule}/{test.test}.{pass_fail}'
+    if test.check == 'sce':
+        return f'sce/{test.rule}/{test.test}.{pass_fail}'
+    else:
+        return f'{test.rule}/{test.test}.{pass_fail}'
 
 
 def unit_tests_from_rules(built_tests_dir, rules, ds):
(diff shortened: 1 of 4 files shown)


per-rule/test.py
```diff
@@ -31,7 +31,10 @@
 
 def format_test(test):
     pass_fail = 'pass' if test.is_pass else 'fail'
-    return f'{test.rule}/{test.test}.{pass_fail}'
+    if test.check == 'sce':
+        return f'sce/{test.rule}/{test.test}.{pass_fail}'
+    else:
+        return f'{test.rule}/{test.test}.{pass_fail}'
```
comps: This works fine, but what about following the example of `pass_fail` from a few lines above and just selectively adding a prefix?

```
prefix = 'sce/' if test.check == 'sce' else ''
return f'{prefix}{test.rule}/{test.test}.{pass_fail}'
```

It would avoid the `'{test.rule}/{test.test}.{pass_fail}'` duplication.

Either works for me, choose per your preference.

comps (commented):
Thanks for taking the time to do this properly. I left one minor comment, otherwise LGTM assuming you've tested this - the logic looks fine to me, but I haven't traced every execution path in my head.

per-rule/test.py
vojtapolasek: Makes sense. I made a commit.

PR #646, opened 2026-08-04
/static-checks/ansible/lint: new test for linting ansible playbooks

matusmarhefka:
Resolves #631

Also tracked in https://redhat.atlassian.net/browse/OPENSCAP-7049

### DIFF
--- a/static-checks/ansible/lint/main.fmf
+++ b/static-checks/ansible/lint/main.fmf
@@ -1,0 +1,26 @@
+summary: Lint Ansible playbooks with ansible-lint.
+description: |-
+    Verify that all generated playbooks and all playbooks provided
+    by the scap-security-guide package pass ansible-lint (profile min).
+    The min profile (which includes only mandatory lint rules) is used
+    instead of basic (or stricter) because CaC/content playbooks are
+    machine-generated and use non-standard formatting which would fail
+    many style-oriented rules.
+test: $CONTEST_PYTHON -m lib.runtest ./test.py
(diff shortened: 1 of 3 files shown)


static-checks/ansible/lint/main.fmf
```diff
@@ -0,0 +1,22 @@
+summary: Lint Ansible playbooks with ansible-lint.
+description: |-
+    Verify that all generated playbooks and all playbooks provided
+    by the scap-security-guide package pass ansible-lint (profile min).
```
comps, line 4: Could this please explain, maybe in a second paragraph, why `--profile min`?

matusmarhefka: I tried `basic` (https://docs.ansible.com/projects/lint/profiles/) but that one generated thousands of errors, most of them related to styles and formatting which I don't expect would be fixed. We could try to come up with a list of checks which would be skipped from `basic` profile and keep only important ones. Wdyt?

comps: I don't think that's necessary, I just wanted something like

> The "min" profile skips various style checks and only includes rigorous logic linting, which is what we want here as CaC/content playbook snippet formatting may be non-standard.

(if this is true, might not be, I just made it up)

Basically, "min" means nothing to me, and I just wanted some details about what it is.

matusmarhefka: Updated.

PR #650, opened 2026-08-11
Always test per-rule playbooks in user content

comps:
This impacts
```
/static-checks/ansible/lint
/static-checks/ansible/syntax-check
```
by making them always test per-rule playbooks:

- if CONTEST_CONTENT is given, the playbooks are explicitly built (or reused if built already)
- else we're testing RPM content and rely on per-rule-playbook RPM to be installed via 'recommend'
  - we could enforce its presence, but we probably want the tests usable even when it's not available, so RPM testing is still opportunistic

### DIFF
--- a/static-checks/ansible/syntax-check/test.py
+++ b/static-checks/ansible/syntax-check/test.py
@@ -5,32 +5,35 @@ import subprocess
 from lib import ansible, util, results
 
 
+def check_playbook(playbook, name):
+    cmd = ['ansible-playbook', '--syntax-check', playbook]
+    ret = util.subprocess_run(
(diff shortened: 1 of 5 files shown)


static-checks/ansible/syntax-check/test.py
```diff
@@ -5,32 +5,36 @@
 from lib import ansible, util, results
 
 
+def check_playbook(playbook, name):
+    cmd = ['ansible-playbook', '--syntax-check', playbook]
+    # ansible-lint writes violations to stdout
```
matusmarhefka, line 10: remove the copy-pasta comment

matusmarhefka (approved):

PR #660, opened 2026-08-20
Remove outdated waivers after 0.1.82 stabilization

jan-cerny:
These waivers currently waive passed tests because the linked issues have been fixed, therefore they can be removed.

### DIFF
--- a/conf/waivers/permanent
+++ b/conf/waivers/permanent
@@ -63,9 +63,6 @@
 /static-checks/html-links/https://www.ccn-cert.cni.es/pdf/guias/series-ccn-stic/guias-de-acceso-publico-ccn-stic/6768-ccn-stic-610a22-perfilado-de-seguridad-red-hat-enterprise-linux-9-0/file.html
     "URL returned error: 403" in note
 # likely has bot detection
-/static-checks/html-links/https://www.cyber.mil/stigs/downloads/.*
-    "Connection reset by peer" in note
-# also bot detection?
 /static-checks/html-links/https://www.cyber.gov.au/acsc/view-all-content/ism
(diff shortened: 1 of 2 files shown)


conf/waivers/permanent
```diff
@@ -63,9 +63,6 @@
 /static-checks/html-links/https://www.ccn-cert.cni.es/pdf/guias/series-ccn-stic/guias-de-acceso-publico-ccn-stic/6768-ccn-stic-610a22-perfilado-de-seguridad-red-hat-enterprise-linux-9-0/file.html
     "URL returned error: 403" in note
 # likely has bot detection
-/static-checks/html-links/https://www.cyber.mil/stigs/downloads/.*
```
matusmarhefka: I would keep these html-links waivers. While the issue did not appear in the current stabilization, it doesn't mean we won't encounter it in the future.

conf/waivers/productization
```diff
@@ -104,22 +101,12 @@
...
 # https://github.com/ComplianceAsCode/content/issues/14570
 /hardening/host-os/oscap/.+/rsyslog_files_permissions
     rhel.is_centos() and rhel in [9, 10]
 
-# https://github.com/ComplianceAsCode/content/issues/14992
-/hardening/host-os/ansible/.+/mount_option_.+
-/hardening/host-os/ansible/.+/systemd_tmp_mount_enabled
-/hardening/container/.+/mount_option_.+
+# https://github.com/ComplianceAsCode/content/issues/15002
+/hardening/host-os/ansible/.+/mount_option_tmp_noexec
```
matusmarhefka: Just FTR, the waiver makes sense, but this result already matches an existing waiver, see https://github.com/RHSecurityCompliance/contest/blob/main/conf/waivers/infrastructure#L9-L11

matusmarhefka (approved):

PR #664, opened 2026-08-26
Enable CUI profile testing on RHEL-10

ggbecker:
## Description

Enables the `/cui` profile test cases on RHEL-10 (and CentOS Stream 10) across the hardening and scanning plans.

The NIST 800-171 (CUI) profile is now available on RHEL-10 via ComplianceAsCode/content#15062. This PR removes the "there is no CUI profile on RHEL-10+" `adjust+` blocks that previously disabled the `/cui` test cases on those distros, so CUI is now exercised on all supported RHEL versions.

### Changes

- 21 `.fmf` files updated, removing the RHEL-10+ CUI exclusions across `hardening/` and `scanning/` plans.
- Where an `adjust+` block held only the RHEL-10 exclusion, the block was removed; where it also carried other conditions (GUI incompatibility on RHEL-8, `rng-tools` / `rhc-worker-playbook` requires, FIPS tags), those were preserved.

### Dependency

Requires the RHEL-10 CUI profile from ComplianceAsCode/content#15062. RHEL-8/9 already ship the CUI profile and run unchanged.


hardening/anaconda/with-gui.fmf
```diff
@@ -48,11 +48,8 @@ tag+:
     tag+:
       - fips
     adjust+:
-      - when: distro == rhel-8 or distro == centos-stream-8
-        enabled: false
-        because: >
-            not supported on RHEL-8 according to RHEL documentation,
-            the "Profiles not compatible with Server with GUI" table
+      - enabled: false
+        because: CUI profile is not supported with GUI
```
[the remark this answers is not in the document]
ggbecker: Yes, this is a change that has to land in sync with https://github.com/ComplianceAsCode/content/pull/15062

We are enabling CUI profile across all RHEL products and that profile is deriving from the NIST-800-53 controls which were primarily based on CIS Level 2 Server benchmark which is a profile targeting non-gui deployments AFAIK.


Looking at the **CIS** definition:
```
/cis:
    adjust+:
      - enabled: false
        because: >
            not supported on RHEL-8 or RHEL-9 according to RHEL documentation,
            the "Profiles not compatible with Server with GUI" table
```

If you think I'm not seeing something, please advise.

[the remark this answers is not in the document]
comps: The section I commented on is a CUI section:

```diff
diff --git a/hardening/anaconda/with-gui.fmf b/hardening/anaconda/with-gui.fmf
index d1207a9..92ecdc3 100644
--- a/hardening/anaconda/with-gui.fmf
+++ b/hardening/anaconda/with-gui.fmf
@@ -45,17 +45,14 @@ tag+:
       - subset-profile
 
 /cui:
     tag+:
       - fips
     adjust+:
-      - when: distro == rhel-8 or distro == centos-stream-8
-        enabled: false
-        because: >
-            not supported on RHEL-8 according to RHEL documentation,
-            the "Profiles not compatible with Server with GUI" table
+      - enabled: false
+        because: CUI profile is not supported with GUI
 
 /e8:
 
 /hipaa:
 
 /ism_o:
```
The previous condition disabled it only for RHEL-8, so it was likely running on RHEL-9 (if it was disabled on 10).\
Your change disables it globally for all RHELs.

Was that intentional?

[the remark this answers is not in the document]
ggbecker: Yeah, that was intentional. CUI now, for all RHEL versions, will be based on CIS L2 Server (fundamentally through the NIST-800-53) so we have to touch all RHEL versions here on whatever decision we make.

comps:
LGTM on first glance, it just needs the `# (tracking issue to be filed)` on Image Builder and a synchronized merge of https://github.com/ComplianceAsCode/content/pull/15062 along with this PR.

Plus rebases of all other upstream PRs (as this test would be failing there).

comps:
Also, before this is merged, we should run a `/plans/weekly` custom pipeline (but `x86_64` only) against it, to catch any non-daily test variants (UEFI, GUI, etc.) and extend waivers to them too, if appropriate.

ggbecker:
> Also, before this is merged, we should run a `/plans/weekly` custom pipeline (but `x86_64` only) against it, to catch any non-daily test variants (UEFI, GUI, etc.) and extend waivers to them too, if appropriate.

Okay, I will run that soon.

comps:
The CI failure is because there's currently no way to specify both CaC/content and Contest PRs, so upstream master was used: https://artifacts.dev.testing-farm.io/e5636026-e456-4ced-b3d2-bd4a82e5beec/

edit: actually, I lied.

/packit test -i productization --env TESTS=/cui$ --env CONTENT_PR=15062

comps:
/packit test -i productization --env TESTS=/cui$ --env CONTENT_PR=15062 --env PLAN=/plans/weekly

comps:
The CentOS Stream tests failed: https://artifacts.dev.testing-farm.io/1e4272f0-c106-4a80-9ad7-af6172ebc753/ but I've seen these failures before too, so they were not introduced by this PR.

ggbecker:
https://github.com/RHSecurityCompliance/contest/pull/664/changes/15741acf99f9fe09c527427dfd663e8a0ee9e12c

Adds waivers for the two remaining issues that do not behave well on containers (image mode). Please see the commit message for details. The conclusion was basically that those rules cannot be remediated in the built image. There were certain conditions on why the error manifested only on RHEL9, it should be described in the commit message as well.

conf/waivers/productization
```diff
@@ -137,4 +137,30 @@
...
+# comparison is skipped. Only RHEL 9 ships CUI in both old and new content.
+#
+# aide_build_database on Image Mode Day-2 upgrades: on "bootc switch" the guest
+# /var is not re-provisioned, so the AIDE database built into the new image never
+# lands on disk. The released (old) CUI profile did not select this rule, so the
+# initial /var seed lacked the database and the switch does not add it (see the
+# rule's Image Mode warning: the AIDE db must be regenerated after each update).
+# Fresh Image Mode (bootc-image-builder, anaconda-ostree) passes; only the
+# old-new upgrade path is affected. Transitional: resolves once a released CUI
+# baseline already includes the rule.
```
comps: "These two waivers" should probably be a single block (connected via `#` comment) as we treat separated waiver blocks as generally different problems.

Also, I don't agree with the reasoning - what happens once we release on RHEL-10? Will the failures appear there too? If they are valid regardless of RHEL version, why scope them *to* a RHEL version?

Also, this is `productization` waivers, so a content issue URL is missing.

conf/waivers/productization
```diff
@@ -137,4 +137,30 @@
...
+/hardening/container/old-new/cui/aide_build_database
+    rhel == 9
+
+# ensure_pam_wheel_group_empty on Image Mode Day-2 upgrades: the su/wheel
+# access group is created imperatively in /etc/group at image build time, but
+# on "bootc switch" /etc is reconciled with a 3-way merge that can drop the
+# build-time group entry, so the post-upgrade scan intermittently fails.
+# Confirmed flaky: with identical inputs it passes on some instances and fails
+# on others. Upstream issue to be filed.
```
comps: > Upstream issue to be filed.

conf/waivers/productization
ggbecker: > "These two waivers" should probably be a single block (connected via `#` comment) as we treat separated waiver blocks as generally different problems.
> 

I will group them into a unified single block.

> Also, I don't agree with the reasoning - what happens once we release on RHEL-10? Will the failures appear there too? If they are valid regardless of RHEL version, why scope them _to_ a RHEL version?

Yeah, you are correct. I believe we should include RHEL10 (and any subsequent versions), in this case it would only keep RHEL8 out of question because it doesn't run the container test.

Nevertheless, it probably would work fine if we run on "all" versions (aka "True" condition).

> 
> Also, this is `productization` waivers, so a content issue URL is missing.

Yeah, that kept me wondering, I believe this is more like a permanent waiver since we cannot "fix" during the image build.

conf/waivers/productization
ggbecker: I'm more inclined into turning this into a permanent waiver. Let me take a look and see if that can be done.

conf/waivers/test_specific
```diff
@@ -0,0 +1,26 @@
...
+# image (hardened with the currently shipped content) is booted, then switched
+# ("bootc switch") to a new image (hardened with the tested CaC content), and
+# scanned post-switch. "bootc switch" does not re-run remediation and does not
+# re-provision /var, and only 3-way-merges /etc against the live system - it
+# does not simply replace guest state with the new image's state. So rules that
+# the *old* content didn't select can fail post-switch even though the new
+# image itself was hardened correctly, because whatever the new image's build
+# produced under /var (or a locally-diverged /etc) never reaches the running
+# guest.
+# ---
```
comps: Could we please just add a TODO here, so somebody notices when these waivers are not needed anymore, because old content will have the rules?

```
#
# TODO: Remove these waivers when the "old" content contains the rules.
# ---
```

It might get picked up by the "unused waivers" logic, but I think it applies only to `productization`, and a TODO will be seen by regular greps we do when filing new GH issues / improvements.

ggbecker: Done

PR #665, opened 2026-08-27
Waive file_permissions_sshd_private_key on image-builder and anaconda

ggbecker:
SSH host keys don't exist at image creation / anaconda install time, so the remediation cannot fix them. They are generated on first boot by sshd-keygen and only then become non-compliant, especially since ComplianceAsCode/content#15056 now requires 0600 even for the ssh_keys group owner while RHEL-generated keys are 0640 root:ssh_keys.


conf/waivers/permanent
```diff
@@ -108,6 +108,17 @@
...
+# image creation / installation, before the SSH host keys exist. The keys are
+# generated on first boot by sshd-keygen and only then become non-compliant, so
+# nothing is remediated at scan time. A second remediation after reboot would fix
+# it, but that's not something we want to do here.
+# On top of that, the referenced pull request now requires 0600 permissions even
+# for the ssh_keys group owner of the private key. Keys generated by RHEL have
+# 0640 root:ssh_keys, so once they are generated (and remediation is not re-run
+# after the first boot) the rule reports a non-compliant state.
+# https://github.com/ComplianceAsCode/content/pull/15056
+/hardening/(image-builder|anaconda)/.+/file_permissions_sshd_private_key
```
ggbecker: Should we actually merge with the previous waiver though?

`/hardening/container/.+/file_permissions_sshd_private_key`

something like:


```suggestion
/hardening/(container|image-builder|anaconda)/.+/file_permissions_sshd_private_key
```

Arden97: We probably should. Just make sure to add an issue link to that waiver as well.

ggbecker: Okay. Regarding the issue, I don't think we should have an actual issue for that. If you see the case above, it's it of the same nature and there isn't an issue there as well. There is nothing to fix there because the keys are generated on the first boot so nothing to be fixed while the image is built.

Arden97 (approved):

